PR #16, opened 2017-12-27
Tweak interfaces

lahma:
Less ambitious version of interface fixes. Sprinkling some AggressiveInlining to small methods too.

Fixes hoisting scope to have concrete List<T> and return SequenceExpression back to having Expression as List's type instead of INode.


lahma:
@sebastienros a new version on NuGet would be also highly appreciated if you find this mergeable.

PR #25, opened 2018-04-25
Add HoistingScope to ArrowFunctions

maximburyak:
Fix incorrect HoistingScope in ArrowFunctions


sebastienros:
I assume you only find the issue when you start using this expression :/

PR #28, opened 2018-06-19
Added a `net45` target

Taritsyn:


lahma:
Now each project has different full framework target. Would it make sense to target net452 in all (with netstandards2.0) which is the oldest that is still being supported? I understand the need as netstandard2.0 is only supported on 4.6.1 onwards.

Taritsyn:
Hello, Marko!

In fact, my purpose is to make sure, that the Jint version 3.X could be used in old projects. At the moment, Jint version 2.11.58 supports the following targets: `net40`, `net45`, `net451`, `netstandard1.3` and `netstandard2.0`.

For example, [AutoMapper](https://www.nuget.org/packages/AutoMapper/), [Autofac](https://www.nuget.org/packages/Autofac/) and [Json.NET](https://www.nuget.org/packages/Newtonsoft.Json/) still support a .NET Framework 4.5.

sebastienros:
I am fine with targeting an older version, like `net45`. If assume we have to test also on `net45` if we want to ship it. However we don't need to change the other projects. And what we should definitely not do is use some many __different__ targets, I can see `net46`, `net461`, `net45` and `net451`.

sebastienros:
Also esprima only supports netstandard1.3. That's something to consider.

sebastienros:
If there is no code difference, then we just need to target the minimum one. This PR doesn't require any changes in the code, or conditional sections.

Taritsyn:
You can close this PR. My purpose is not to make a PR, but add support for the `net40`, `net45`, `netstandard1.3` targets to the Jint version 3.X, so that it can be used in the [JavaScript Engine Switcher](https://github.com/Taritsyn/JavaScriptEngineSwitcher) version 3.X.

lahma:
Microsoft is advocating of using netstandard2.0 and forgetting the rest. So I'd suggest net452 netstandard2.0.

sebastienros:
And we don't need to change the Benchmarks, we won't look at it anyway.

Taritsyn:
> Microsoft is advocating of using netstandard2.0 and forgetting the rest. So I'd suggest net452 netstandard2.0.

I think it only applies to ASP.NET Core, because the `System.*` packages are still compatible with a .NET Framework 4.5 and .NET Standard 1.X (for example, [System.Text.Encodings.Web](https://www.nuget.org/packages/System.Text.Encodings.Web/)).

PR #32, opened 2018-08-15
Fix Symbol.iterator static member as object property key

lahma:
I'm not entirely sure if I got this right, but found about this problem when checking Test262 cases and Esprima crashed for the given sample script.


sebastienros:
Note to myself: find the fix in Esprima and check the change. And also apply all other fixes.

sebastienros:
Fixed

PR #43, opened 2018-09-27
Fix for Esprima.Sample System.IO.FileNotFoundException ... Newtonsoft.Json

JarLob:
I tried to run Esprima.Sample and got System.IO.FileNotFoundException. I also couldn't debug it. I tried different things until I noticed the project builds Esprima.Sample.dll instead of Esprima.Sample.exe (Weirdly
enough the Main function was executed).

### DIFF
--- a/samples/Esprima.Sample/Esprima.Sample.csproj
+++ b/samples/Esprima.Sample/Esprima.Sample.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


JarLob:
Hmm, I can swear it built .exe. But now I see .dll again. So the comment about .exe vs .dll maybe somewhat misleading. But the commit is good. It fixes the exception and debugging issue...

sebastienros:
You can't have an exe if you don't do a self contained deployment.
And `dotnet foo.dll` is how it works
I merged only because it makes sense to target netcoreapp

PR #51, opened 2018-12-04
Convert HoistingScope properties to fields

lahma:
Even these two getters show up in performance profiling for script-intensive scenarios. Let's change to direct field access.


KvanTTT:
Have you tried RELEASE mode?

lahma:
Yep. I'm running the profiled app in release mode, which uses Jint built in release mode, which references Esprima from NuGet where it should be [built in release mode](https://github.com/sebastienros/esprima-dotnet/blob/dev/appveyor.yml#L11). There's no guarantee such getter gets inlined.

PR #62, opened 2019-01-03
Swap ThrowError with CreateError as IErrorHandler extension

atifaziz:
This PR swaps `CreateError` with `ThrowError` as an extension of `IErrorHandler`.

## Rationale

`CreateError` should be a member of the `IErrorHandler` interface and a decision of the implementation and `ThrowError` is the mere boilerplate and can (practically) never vary, so it's more naturally expressed as an extension. In fact, I think `ThrowError` shouldn't exist because it interferes with the C# compiler's ability to do control and data flow analysis. That's a different story and if it makes sense, I can retire it in a separate PR.

### DIFF
--- a/src/Esprima/ErrorHandlerExtensions.cs
+++ b/src/Esprima/ErrorHandlerExtensions.cs
@@ -2,18 +2,9 @@
 {
     public static class ErrorHandlerExtensions
     {
-        public static ParserException CreateError(this IErrorHandler handler, int index, int line, int col, string description)
+        public static void ThrowError(this IErrorHandler handler, int index, int line, int column, string message)
         {
-            var msg = $"Line {line}': {description}";
-            var error = new ParserException(msg)
-            {
-                Index = index,
-                Column = col,
-                LineNumber = line,
-                Description = description,
-                Source = handler.Source
-            };
-            return error;
+            throw handler.CreateError(index, line, column, message);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Esprima/ErrorHandlerExtensions.cs
```diff
@@ -2,18 +2,9 @@
...
-            var error = new ParserException(msg)
-            {
-                Index = index,
-                Column = col,
-                LineNumber = line,
-                Description = description,
-                Source = handler.Source
-            };
-            return error;
+            throw handler.CreateError(index, line, column, message);
```
lahma: could someone choose not to create exception and return null, causing this to throw null?

atifaziz: Good point. I see one of two ways to approach that: (a) be super paranoid and check against undefined behaviour or (b) leave it as _undefined behaviour_ (albeit undocumented 😞 unless we add doc comments). Then again, we don't check that whether the `handler` is null. :wink:

atifaziz: @lahma said (on Gitter chat):

> the create error comment was mostly for whether I was unsure if one would use the API to suppress errors and returning null would be the only way (which wouldn't work)

Seems that [Esprima doesn't allow `null` to be returned](https://github.com/jquery/esprima/blob/c0dfa42da4da9fa242dac623278a082cb38edc20/src/error-handler.ts#L49):

```js
createError(index: number, line: number, col: number, description: string): Error
```

In other words, the return type is not `Error | null` and I'm assuming they compile with `--strictNullChecks` but haven't verified.

Issue #38, opened 2018-09-20
Implement convenient ToString() methods for nodes and locations


KvanTTT:
Or use `DebuggerDisplay` attribute instead. It makes the debug process more clear.

Also, I suggest using the following output line-column based format for `Location`:

* `[l,c)` - if start position equals to the end position.
* `[l,c1..c2)` - if start line equals to the end line, but columns are different.
* `[l1..l2,c)` - if start line differs from the end, but columns are the same.
* `[l1,c1..l2,c2)` - common format.

And the following for index-based (Roslyn-like):

* `[i)` - if start position equals to the end position.
* `[i1..i2)` - includes the first position but excludes the latest one.

sebastienros:
I am fine with that. Do you want to create a PR for this?

KvanTTT:
Yes, I want, but later.

KvanTTT:
What is the reason for using array ranges `int[] Range` instead of typed ranges like `TextSpan Range` where `TextSpan` consist of `Start` and `Length` or `Start` and `Stop` fields? I think it's more convenient.

PR #65, opened 2019-01-07
Make Range a more wholesome implementation

atifaziz:
This PR makes `Range` a more wholesome implementation that you'd expect from a public value type by adding the following:

- Validation of start and end at construction
- Implementation for `object` members (`GetHashCode`, `Equals` & `ToString`).
- Support for `IEquatable<Range>`
- Implementation for equality operators (`==` and `!=`).

### DIFF
--- a/src/Esprima/Ast/Range.cs
+++ b/src/Esprima/Ast/Range.cs
@@ -1,14 +1,32 @@
 ﻿namespace Esprima.Ast
 {
-    public struct Range
+    using System;
+    using System.Globalization;
+
+    public readonly struct Range : IEquatable<Range>
     {
         public readonly int Start;
         public readonly int End;
 
         public Range(int start, int end)
         {
-            Start = start;
-            End = end;
+            Start = start >= 0 ? start : throw new ArgumentOutOfRangeException(nameof(start));
+            End = end > start ? end : throw new ArgumentOutOfRangeException(nameof(end));
         }
+
+        public bool Equals(Range other) =>
+            Start == other.Start && End == other.End;
+
+        public override bool Equals(object obj) =>
+            obj is Range other && Equals(other);
+
+        public override int GetHashCode() =>
+            unchecked((Start * 397) ^ End);
+
+        public override string ToString() =>
+            string.Format(CultureInfo.InvariantCulture, "{0}...{1}", Start, End);
+
+        public static bool operator ==(Range left, Range right) => left.Equals(right);
+        public static bool operator !=(Range left, Range right) => !left.Equals(right);
     }
 }


src/Esprima/Ast/Range.cs
```diff
@@ -1,14 +1,32 @@
...
+            Start == other.Start && End == other.End;
+
+        public override bool Equals(object obj) =>
+            obj is Range other && Equals(other);
+
+        public override int GetHashCode() =>
+            unchecked((Start * 397) ^ End);
+
+        public override string ToString() =>
+            string.Format(CultureInfo.InvariantCulture, "{0}...{1}", Start, End);
```
KvanTTT: I suggest using the following format: `[start..length)` (include the left border and exclude the right one).

sebastienros:
Where is the text representation used?

atifaziz:
> Where is the text representation used?

Nowhere but was done for sake of completeness, because it was cheap and simple to do. If anything for now, it does help with watches when debugging.

atifaziz:
@lahma I thought of `DebuggerDisplay` but that's a different problem altogether. The string representation proposed here is a non-binding contract. There is no `Range` parsing method and hopefully never going to be one because, well, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), right? `$"{Start}...{End}"` therefore seems like a reasonable thing to have. You get two benefits for free: the debugger uses it and if someone prints it to the console then they get something better than `Esprima.Ast.Range` everywhere.

If there is some debate about `$"{Start}...{End}"`, `$"{Start}..{End}"` or `$"[{Start}..{End})"` then perhaps we can just go for `$"Range({Start}, {End})`. Then there's no ambiguity as it mimics reflects the construction.

lahma:
I personally like `$"Range({Start}, {End})"`, easy to reason with.

src/Esprima/Ast/Range.cs
KvanTTT: I suggested such text span because of Roslyn compiler uses [such format for TextSpan](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Text/TextSpan.cs#L239) and I found it convenient and clear.

KvanTTT:
I suggested such text span because of Roslyn compiler uses [such format for TextSpan](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Text/TextSpan.cs#L239) and I had found it convenient, formal and clear.

KvanTTT:
BTW, there is the related issue https://github.com/sebastienros/esprima-dotnet/issues/38

lahma:
Thanks for linking the issue, following Roslyn pattern of course has its merits too.

atifaziz:
> BTW, there is the related issue #38

Didn't realise but thanks for the pointer. This PR definitely partially addresses #38.

> I suggested such text span output because of Roslyn compiler uses [such one](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Text/TextSpan.cs#L239) and I had found it convenient, formal and clear.

I wasn't aware but this is good to know.

> following Roslyn pattern of course has its merits too.

Right and what's important here is that we've had a thought-process and rationale for each of the proposals:

1. `start...end`: bracket-less interval notation but which is Ruby-inspired for three-dots to means excluding the end
2. `[start..end)`: Uncommon but familiar in dependency version specification domain and used by Roslyn
3. `Range(start, end)`: Unambiguous and self-explanatory though arguably uncommon to see type name in an overridden implementation of `Object.ToString`.

I'm happy with settling with 2. If someone wonders if `)` was a typo, :wink: there's a thread of thoughts to follow here.

lahma:
Option 2 indeed sounds like the way to go. Option 3 might be good candidate for DebuggerDisplay as it's then closer to code when you are running the debugger.

PR #67, opened 2019-01-08
Make methods that can be static, static

atifaziz:
The title says it all.

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -286,7 +286,7 @@ namespace Esprima
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private MetaNode StartNode(Token token)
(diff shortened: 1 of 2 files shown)


src/Esprima/JavascriptParser.cs
```diff
@@ -3451,7 +3451,7 @@ private List<StatementListItem> ParseDirectivePrologues()
 
         // https://tc39.github.io/ecma262/#sec-method-definitions
 
-        private bool QualifiedPropertyName(Token token)
+        private static bool QualifiedPropertyName(Token token)
         {
             switch (token.Type)
```
lahma: mental not for me, need to check how this inlines as uses `switch` instead of `if`s, also probably could be simpler (and probably less obvious) via `<` `>` operators

atifaziz:
@sebastienros I imagine this is a quick one?

Issue #66, opened 2019-01-07
AST API has inconsistencies


atifaziz:
The AST API currently contains a few inconsistencies, fixing which would go a long way to polish things up for 1.0.

The majority of the AST appears to be read-only except in a few select cases, most of which seem to be unused (or are/were supporting Jint?). The biggest read-write structure seems to be `List<>` but `List<>` is not designed to be exposed in the public API surface of a class library anyway. I suggest rendering the entire public AST data structures read-only for sake of consistency, if not for other benefits that may stem from that like naturally thread-safe structures due to their immutability.

There is also mixed use of fields and properties. It would seem best to consistently expose properties.

The table below shows all node classes of the AST with fields and properties. It shows the mixed use of the two. I've also highlighted read-write properties/fields with an asterisk (*), which would need review to make the AST immutable.

| Type | Fields | Properties |
|------|--------|------------|
| `ArrayExpression` | `Elements` |  |
| `ArrayPattern` | `Elements` |  |
| `ArrowFunctionExpression` | `Id`, `Params`, `Body`, `Generator`, `Expression`, `HoistingScope`\* |  |
| `ArrowParameterPlaceHolder` | `Params` |  |
| `AssignmentExpression` | `Operator`, `Left`, `Right` |  |
| `AssignmentPattern` | `Left`, `Right`\* |  |
| `BinaryExpression` | `Operator`, `Left`, `Right` |  |
| `BlockStatement` | `Body` |  |
| `BreakStatement` | `Label` |  |
| `CallExpression` | `Callee`, `Arguments`, `Cached`\*, `CanBeCached`\*, `CachedArguments`\* |  |
| `CatchClause` | `Param`, `Body` |  |
| `ClassBody` | `Body` |  |
| `ClassDeclaration` | `Id`, `SuperClass`, `Body` |  |
| `ClassExpression` | `Id`, `SuperClass`, `Body` |  |
| `ConditionalExpression` | `Test`, `Consequent`, `Alternate` |  |
| `ContinueStatement` | `Label` |  |
| `Directive` | `Directiv` |  |
| `DoWhileStatement` | `Body`, `Test` |  |
| `ExportAllDeclaration` | `Source` |  |
| `ExportDefaultDeclaration` | `Declaration` |  |
| `ExportNamedDeclaration` | `Declaration`, `Specifiers`, `Source` |  |
| `ExportSpecifier` | `Exported`, `Local` |  |
| `ExportStatement` | `Expression` |  |
[...]

PR #68, opened 2019-01-08
Replace BCL list for AST with a publicly read-only list

atifaziz:
This PR partially addresses issue #66. It moves in the direction of hopefully making the AST immutable. It does this by replacing `System.Collections.Generic.List<>` with a local implementation for the following reasons:

- BCL's `List<>` is not designed to be exposed from the public API of a library
- `List<>` proposed here:
  - has an _externally/publicly immutable_ API
  - is a `struct` so it discounts heap allocations as a mere wrapper for a dynamically managed array

Benchmarks before this PR:

| Method         | FileName              |   Mean |  Error | StdDev |    Gen 0 |    Gen 1 |  Gen 2 | Allocated |
|----------------|-----------------------|-------:|-------:|-------:|---------:|---------:|-------:|----------:|
| `ParseProgram` | `angular-1.2.5`       | 29.599 |  0.513 | 0.4548 |   1937.5 |      750 |    250 |     10.02 |
| `ParseProgram` | `backbone-1.1.0`      |  2.325 | 0.0244 | 0.0217 |   281.25 | 136.7188 |        |      1.54 |
| `ParseProgram` | `jquery-1.9.1`        | 21.158 | 0.4142 | 0.3875 |  1593.75 |      625 | 156.25 |      8.52 |
| `ParseProgram` | `jquery.mobile-1.4.2` | 34.655 | 0.6873 | 0.6429 |   2437.5 |    937.5 |    250 |     13.07 |
| `ParseProgram` | `mootools-1.4.5`      | 15.817 | 0.2352 | 0.2085 |     1250 |    562.5 |   62.5 |      7.03 |
| `ParseProgram` | `underscore-1.5.2`    |  1.758 | 0.0126 | 0.0118 | 316.4063 |  25.3906 |        |      1.34 |
| `ParseProgram` | `yui-3.12.0`          | 12.874 | 0.1814 | 0.1697 |   1062.5 |   531.25 |        |      6.31 |

Benchmarks after this PR:

| Method         | FileName              | Mean   | Error  | StdDev | Gen 0    | Gen 1   | Gen 2  | Allocated |
|----------------|-----------------------|-------:|-------:|-------:|---------:|---------:|-------:|----------:|
| `ParseProgram` | `angular-1.2.5`       | 26.217 | 0.2936 | 0.2603 | 1750     | 593.75  | 125    | 9.55      |
| `ParseProgram` | `backbone-1.1.0`      | 2.31   | 0.0261 | 0.0231 | 257.8125 | 125     |        | 1.48      |
| `ParseProgram` | `jquery-1.9.1`        | 17.602 | 0.2442 | 0.2284 | 1468.75  | 750     | 62.5   | 8.18      |
| `ParseProgram` | `jquery.mobile-1.4.2` | 32.568 | 0.2921 | 0.2439 | 2250     | 875     | 187.5  | 12.56     |
| `ParseProgram` | `mootools-1.4.5`      | 13.328 | 0.2587 | 0.2656 | 1156.25  | 578.125 | 15.625 | 6.72      |
| `ParseProgram` | `underscore-1.5.2`    | 1.916  | 0.0096 | 0.0085 | 273.4375 | 89.8438 |        | 1.27      |
| `ParseProgram` | `yui-3.12.0`          | 12.581 | 0.1301 | 0.1154 | 1015.625 | 500     |        | 6.11      |

Below is a comparison of the above two tables:

| FileName              | Mean | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-----------------------|-----:|------:|------:|------:|----------:|
| `angular-1.2.5`       |  11% |   10% |   21% |   50% |        5% |
| `backbone-1.1.0`      |   1% |    8% |    9% |       |        4% |
| `jquery-1.9.1`        |  17% |    8% |  -20% |   60% |        4% |
| `jquery.mobile-1.4.2` |   6% |    8% |    7% |   25% |        4% |
| `mootools-1.4.5`      |  16% |    8% |   -3% |   75% |        4% |
| `underscore-1.5.2`    |  -9% |   14% | -254% |       |        5% |
| `yui-3.12.0`          |   2% |    4% |    6% |       |        3% |

Positive percentages show the differential improvement; negatives show degradation.

Using a `struct` for a list shows an overall improvement in speed (6.3% better on average) and in allocations (4.2% better on average). For now, however, I can't explain why [Underscore](https://github.com/sebastienros/esprima-dotnet/blob/808dc9074091f8027a9dc10e9696b7357557f672/test/Esprima.Tests/Fixtures/3rdparty/underscore-1.5.2.js) is a bit of an odd ball. Would appreciate if someone would like to dig in there a bit and share some findings.

### DIFF
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -16,8 +16,8 @@ namespace Esprima.Ast
 
     public struct List<T> : IReadOnlyList<T>
     {
-        T[] _items;
-        int _count;
+        private T[] _items;
+        private int _count;
 
         internal List(int capacity)
         {
@@ -37,7 +37,7 @@ namespace Esprima.Ast
             _count = list.Count;
         }
 
-        int Capacity => _items?.Length ?? 0;
+        private int Capacity => _items?.Length ?? 0;
 
         public int Count
         {
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -20,7 +20,7 @@ namespace Esprima.Ast
 
         internal List(int capacity)
         {
-            _items = new T[capacity];
+            _items = capacity == 0 ? null : new T[capacity];
             Count = 0;
         }
 
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -4,7 +4,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
-namespace Esprima
+namespace Esprima.Ast
 {
     // This structure is like List<> from the BCL except it is designed
     // to be modifiable by this library during the construction of the AST
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -118,16 +118,6 @@ namespace Esprima.Ast
             Array.Copy(_items, index + 1, _items, index, Count - index);
         }
 
-        public IEnumerator<T> GetEnumerator()
-        {
-            for (var i = 0; i < Count; i++)
-            {
-                yield return _items[i];
-            }
-        }
-
-        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
-
         public T this[int index]
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -156,6 +146,83 @@ namespace Esprima.Ast
             var last = this[lastIndex];
             RemoveAt(lastIndex);
             return last;
+        }
+
+        public Enumerator GetEnumerator() => new Enumerator(_items, Count);
+
+        IEnumerator<T> IEnumerable<T>.GetEnumerator() => GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        /// <remarks>
+        /// This implementation does not detect changes to the list
+        /// during iteration and therefore the behaviour is undefined
+        /// under those conditions.
+        /// </remarks>
+
+        public struct Enumerator : IEnumerator<T>
+        {
+            private int _index;
+            private T[] _items; // Usually null when count is zero
+            private int _count;
+
+            internal Enumerator(T[] items, int count) : this()
+            {
+                _index = -1;
+                _items = items;
+                _count = count;
+            }
+
+            // Since the items can be null when count is zero, a negative
+            // count is used to designate the disposed state.
+
+            private bool IsDisposed => _count < 0;
+
+            public void Dispose()
+            {
+                _items = null;
+                _count = -1;
+            }
+
+            public bool MoveNext()
+            {
+                ThrowIfDisposed();
+
+                if (_index + 1 == _count)
+                {
+                    return false;
+                }
+
+                _index++;
+                return true;
+            }
+
+            public void Reset()
+            {
+                ThrowIfDisposed();
+                _index = -1;
+            }
+
+            public T Current
+            {
+                get
+                {
+                    ThrowIfDisposed();
+                    return _index >= 0
+                         ? _items[_index]
+                         : Throw<T>(new InvalidOperationException());
+                }
+            }
+
+            object IEnumerator.Current => Current;
+
+            private void ThrowIfDisposed()
+            {
+                if (IsDisposed)
+                {
+                    Throw<T>(new ObjectDisposedException(GetType().Name));
+                }
+            }
         }
     }
 }
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using static Esprima.JitHelper;
 
 namespace Esprima.Ast
 {
@@ -123,9 +124,10 @@ namespace Esprima.Ast
 
         public T this[int index]
         {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => index >= 0 && index < Count
                  ? _items[index]
-                 : throw new IndexOutOfRangeException();
+                 : Throw<T>(new IndexOutOfRangeException());
 
             internal set
             {
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -17,7 +17,7 @@ namespace Esprima.Ast
     {
         T[] _items;
 
-        public List(int capacity)
+        internal List(int capacity)
         {
             _items = new T[capacity];
             Count = 0;
--- a/src/Esprima/Ast/List.cs
+++ b/src/Esprima/Ast/List.cs
@@ -17,11 +17,12 @@ namespace Esprima.Ast
     public struct List<T> : IReadOnlyList<T>
     {
         T[] _items;
+        int _count;
 
         internal List(int capacity)
         {
             _items = capacity == 0 ? null : new T[capacity];
-            Count = 0;
+            _count = 0;
         }
 
         public List(List<T> list) : this()
@@ -33,12 +34,16 @@ namespace Esprima.Ast
 
             _items = new T[list.Count];
             list._items.CopyTo(_items, 0);
-            Count = list.Count;
+            _count = list.Count;
         }
 
         int Capacity => _items?.Length ?? 0;
 
-        public int Count { get; private set; }
+        public int Count
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => _count;
+        }
 
         internal List<TResult> Select<TResult>(Func<T, TResult> selector)
         {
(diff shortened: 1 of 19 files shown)


src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+namespace Esprima.Ast
+{
+    // This structure is like List<> from the BCL except it is designed
+    // to be modifiable by this library during the construction of the AST
+    // but publicly read-only thereafter. The only allocation required on the
+    // heap is the backing array storage for the element. An empty list,
+    // however causes no heap allocation; that is, the array is allocated
+    // on first addition.
+
+    public struct List<T> : IReadOnlyList<T>
```
lahma: Maybe this should be called NodeList/AstList/similar so there's clear distinction between BCL one and this?

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            _items = new T[list.Count];
+            list._items.CopyTo(_items, 0);
+            Count = list.Count;
+        }
+
+        int Capacity => _items?.Length ?? 0;
+
+        public int Count { get; private set; }
```
lahma: if count were explicit field could ensure `[MethodImpl(MethodImplOptions.AggressiveInlining)] get => _count;` and of course could directly read/write the field

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+    // but publicly read-only thereafter. The only allocation required on the
+    // heap is the backing array storage for the element. An empty list,
+    // however causes no heap allocation; that is, the array is allocated
+    // on first addition.
+
+    public struct List<T> : IReadOnlyList<T>
+    {
+        T[] _items;
+
+        public List(int capacity)
```
lahma: should there be explicit no-arg constructor and/or otherwise detect 0 capacity => Array.Empty<T>?

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                return;
+            }
+
+            Array.Copy(_items, index + 1, _items, index, Count - index);
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            for (var i = 0; i < Count; i++)
```
lahma: probably causes extra bounds checks as count and _items reference aren't local

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                yield return _items[i];
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public T this[int index]
+        {
+            get => index >= 0 && index < Count
```
lahma: there might still be extra unnecessary bound checks

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Esprima.Ast
+{
+    // This structure is like List<> from the BCL except it is designed
```
lahma: nit: should this be proper `/// <summary>` kind of comment?

src/Esprima/Ast/List.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public T this[int index]
+        {
+            get => index >= 0 && index < Count
+                 ? _items[index]
+                 : throw new IndexOutOfRangeException();
```
lahma: should extract throwing of exceptions to separate method, throwing exception prohibits JIT from inlining

lahma:
Gave some review comments.

> For now, however, I can't explain why underscore is a bit of an odd ball. Would appreciate if someone would like to dig in there a bit and share some findings.

I can try and have a peek a bit later...

src/Esprima/Ast/List.cs
atifaziz: I completely agree. In fact, I completely forgot to say in the opening comments that I'm not happy with the name but it's odd to call it a `NodeList` when `T` is unconstrained.

src/Esprima/Ast/List.cs
atifaziz: > should there be explicit no-arg constructor

A `struct` always has one.

This constructor should in fact be internal because it doesn't make sense for anyone else because they cannot do anything with that initial capacity. I'll fix this.

src/Esprima/Ast/List.cs
atifaziz: It's note for library authors and immaterial for the public. A public summary would just say it's a list/collection of `T`.

src/Esprima/Ast/List.cs
atifaziz: Ah, sorry I misunderstood you and that's a fair point. I'll fix that.

atifaziz: Addressed with 3ad476bd9ca527da712f549c18dc39d45a03a58f.

src/Esprima/Ast/List.cs
atifaziz: Gosh, is the JIT really that dumb? 😒

lahma: Heh, maybe I'd say it's constrained, by time and benefit 🙂 [Some seem to think so at least](https://github.com/ravendb/ravendb/blob/d52e7116bd6e191527b7b5ad9781d626429aa6b6/src/Voron/Data/Fixed/FixedSizeTreePage.cs#L37)

src/Esprima/Ast/List.cs
atifaziz: I'm afraid that won't work as the model stands right now, unless the relations are not all completely thought out. Just to cite one example, `BlockStatement.Body` is a list of `StatementListItem` objects where the `StatementListItem` _interface_ has no relation to `INode`. There many other such instances.

src/Esprima/Ast/List.cs
atifaziz: 236f9682fd60e080bccbf41784d362b28b9fe567 should take care of this.

lahma:
I couldn't yet figure out exact cause why one specific parsing would suffer a bit, but I was able to reproduce it. I don't think it's a showstopper for now. Out of curiosity I compared BCL List against "AstList", [here's the benchmark and results](https://gist.github.com/lahma/156bc7e20de19b881a6fa40342d2d365). 

The biggest takeaways:

* this was testing large lists of 10 000 indices, which I think are a bit uncommon in Esprima parsing
* `foreach` could be faster (custom iterator, no bound checking), but using `for` circumvents the problem (I use for for BCL list too when performance critical)
* `Add` could be faster, but this can be tuned later too

So LGTM - maybe still need to invent better name for the list, @sebastienros can do the authoritative review 😉

atifaziz:
Thanks for running those benchmarks and sharing them.

I had already thought of adding a custom iterator implementation to avoid boxing on `foreach` and started the implementation. Will polish up and share soon.

I wonder if we can do the rename in a separate PR if someone comes up with a good suggestion.

atifaziz:
> - `foreach` could be faster (custom iterator…

Added in 6d00abbe55e401a73203208f431c933b49e0cea6.

src/Esprima/JavascriptParser.cs
```diff
@@ -1192,7 +1192,7 @@ private INode ParseGroupExpression()
                                 if (expr.Type == Nodes.SequenceExpression)
                                 {
                                     var sequenceExpression = expr.As<SequenceExpression>();
-                                    var reinterpretedExpressions = new List<Expression>();
+                                    var reinterpretedExpressions = new Ast.List<Expression>();
```
lahma: I did notice couple of places where the list size is known beforehand like here, so it could be supplied to constructor. But it's an existing problem for sure.

atifaziz:
> I couldn't yet figure out exact cause why one specific parsing would suffer a bit, but I was able to reproduce it.

@lahma So I did a quick dig using the Diagnostic Tools of the VS debugger and I'm not seeing any surprises. Here was the simple test code:

```c#
var source = File.ReadAllText(@"test\Esprima.Tests\Fixtures\3rdparty\underscore-1.5.2.js");
var parser = new JavaScriptParser(source);
parser.ParseProgram();
```

I took a snapshot of the GC heap on line 2 and then on line 3. Here's what the heap looked for before this PR:

image

After this PR, and as expected, the heap has fewer objects and a smaller size:

image

If we look at the top allocated objects before this PR, we can see entries for `System.Collections.Generic.List<>`:

image

…and which are, again as expected, completely gone after applying this PR:

image

So I am still puzzled as to why BDN stats paint another picture, especially with respect to a drastic increase in Gen 1. Anyway, I thought I'd share my findings.

---
PS I only tested against .NET Framework 4.6.1.

atifaziz:
> Conflict

Resolved in 97025591821c8c5a93fe204b43951e63acede727.

PR #71, opened 2019-01-10
Make Node.Type read-only

atifaziz:
This PR changes `Node.Type` to be read-only (in advancement of #66). It did not need to be read-write. It was only set by sub-classes so this PR introduces a protected constructor in `Node` that is then used by subclasses to initialize the node's type.

Making `Node.Type` abstract would have been another option and would have reduced the size of most nodes by one field but I am guessing you want to avoid virtual members for performance reasons so initialization via a protected constructor seemed to be the right thing to do here. Yet another option was to make `Type` setter protected but this wouldn't have enforced initialization of `Type` and leaving it open to modification (even if by subclasses) seemed unnecessary.

---
All tests are passing. ✅

### DIFF
--- a/src/Esprima/Ast/ArrayExpression.cs
+++ b/src/Esprima/Ast/ArrayExpression.cs
@@ -6,9 +6,9 @@ namespace Esprima.Ast
     {
         public readonly List<ArrayExpressionElement> Elements;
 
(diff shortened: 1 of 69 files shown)


atifaziz:
> Please fix the conflict

Fixed in 257955899407b51f9485e2dbea9cf0008f628fcd.

PR #83, opened 2019-01-17
Fix CS0649 warnings about unused fields

atifaziz:
Fixes following warnings:

    JavascriptParser.cs(30,31): warning CS0649: Field 'JavaScriptParser.Context.MarkerStack' is never assigned to, and will always have its default value null
    JavascriptParser.cs(19,24): warning CS0649: Field 'JavaScriptParser.Context.LastCommentStart' is never assigned to, and will always have its default value 0

[These can be found in the last CI build](https://ci.appveyor.com/project/SebastienRos/esprima-dotnet/builds/21672606#L57).

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -16,7 +16,6 @@ namespace Esprima
 
         private sealed class Context
         {
-            public int LastCommentStart;
             public bool AllowIn;
             public bool AllowYield;
             public Token FirstCoverInitializedNameError;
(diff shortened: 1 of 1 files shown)


atifaziz:
@sebastienros Odd these are never used no? Safe to delete? We can always revert and bring them back if there was oversight in the port from the JS version.

sebastienros:
It's part of some work that still needs to be done. I haven't integrated this part from esprima yet.

atifaziz:
I see. Do we then want to leave these warnings or remove them for now and revert when you get round to integrating?

sebastienros:
You can remove them

atifaziz:
[Build is not getting cleaner](https://ci.appveyor.com/project/SebastienRos/esprima-dotnet/builds/21705379). ✨

PR #89, opened 2019-01-23
Redundant BinaryExpression alloc in parser

atifaziz:
The `expr` variable and allocation is not used anywhere and possibly redundant with line 1685 later:

https://github.com/sebastienros/esprima-dotnet/blob/e92de0250c9cfb626541cb49fbd1e2f47b68bcb9/src/Esprima/Parser/JavascriptParser.cs#L1685

Removing it, which this PR suggests, doesn't break tests. Meanwhile, represents wasteful allocation and work.

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -1675,7 +1675,6 @@ namespace Esprima
                         right = (Expression)stack.Pop();
                         var op = ((Token)stack.Pop()).Value;
                         left = (Expression)stack.Pop();
-                        expr = new BinaryExpression((string)op, left, right);
 
                         markers.Pop();
                         var node = StartNode(markers.Peek());


atifaziz:
For reference, the [original implementation doesn't have that extra binary expression allocation](https://github.com/jquery/esprima/blob/c0dfa42da4da9fa242dac623278a082cb38edc20/src/parser.ts#L1512-L1521):

```ts
                // Reduce: make a binary expression from the three topmost entries.
                while ((stack.length > 2) && (prec <= precedences[precedences.length - 1])) {
                    right = stack.pop();
                    const operator = stack.pop();
                    precedences.pop();
                    left = stack.pop();
                    markers.pop();
                    const node = this.startNode(markers[markers.length - 1]);
                    stack.push(this.finalize(node, new Node.BinaryExpression(operator, left, right)));
                }
```

PR #91, opened 2019-02-12
Tidy up arrow functions

AmateurMemetics:
ArrowFunctionExpression needs to be an IFunction to share an interface with other function types.

IFunction has been changed to use INode for the function body, so that it can be either a BlockStatement or an Expression.

Tests have been kept essentially the same.

Also added a small bugfix that EndColumn can be equal to StartColumn on a file which has multiple lines which are all comments.

### DIFF
--- a/src/Esprima/Ast/ArrowFunctionExpression.cs
+++ b/src/Esprima/Ast/ArrowFunctionExpression.cs
@@ -2,15 +2,21 @@
 
 namespace Esprima.Ast
 {
-    public class ArrowFunctionExpression : Node, Expression
+    public class ArrowFunctionExpression : Node, Expression, IFunction
     {
-        public readonly Identifier Id;
-        public readonly List<INode> Params;
-        public readonly INode Body; // : BlockStatement | Expression;
-        public readonly bool Generator;
-        public readonly bool Expression;
+        public bool Strict { get; }
 
-        public HoistingScope HoistingScope;
+        public Identifier Id { get; }
+
+        public List<INode> Params { get; }
+
+        public INode Body { get; } // : BlockStatement | Expression;
+
+        public bool Generator { get; }
+
+        public bool Expression { get; }
+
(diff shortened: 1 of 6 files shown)


AmateurMemetics:
Thanks for the reply. In my haste to get the tests for Jint working I didn't re-run the test suite on Esprima, I'll take that change out of this PR today and raise it separately with a test.

sebastienros:
Can you explain the issue it's fixing?

AmateurMemetics:
Hi Sébastien, this PR allows ArrowFunctionExpression to implement IFunction. Whilst that doesn't explicitly change how Esprima parses, it makes it far simpler to implement arrow functions in Jint (JintFunctionInstance needs an IFunction), which to me makes the most sense since an arrow function can be used anywhere that a normal function can be.

sebastienros:
And why changing `IFunction` ? isn't `BlockStatement` the only think that can get it? Can you share an example otherwise ?

AmateurMemetics:
Arrow functions can have an `Expression` instead of a `BlockStatement`

sebastienros:
Example: http://esprima.org/demo/parse.html?code=material%20%3D%3E%20material.length%0A

AmateurMemetics:
Unless I'm mistaken, the example you gave already demonstrates that the Body can be an Expression which isn't a BlockStatement

sebastienros:
Yes, exactly, I found the answer myself, super proud ;)

sebastienros:
Anyone knows why these tests don't even catch this up?

https://github.com/sebastienros/esprima-dotnet/tree/dev/test/Esprima.Tests/Fixtures/ES6/arrow-function

AmateurMemetics:
Unless I'm mistaken again, that would be because ArrowFunctionExpression did have all the correct fields so it was serialised correctly even though it didn't implement IFunction

PR #97, opened 2019-03-01
Fix hoisting scope state transfer and array pattern parsing

lahma:
Fixed couple issues found while integrating latest dev package to Jint.

* spread tried to cast to Expression instead of IArrayPatternElement
* hoisting scope came with empty lists as stack Peek and Pop returned copies, replaced struct with class

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -27,7 +27,7 @@ namespace Esprima
             public HashSet<string> LabelSet;
         }
 
-        private struct HoistingScopeLists
+        private class HoistingScopeLists
         {
             public ArrayList<FunctionDeclaration> FunctionDeclarations;
             public ArrayList<VariableDeclaration> VariableDeclarations;
@@ -1051,7 +1051,7 @@ namespace Esprima
                     break;
                 case Nodes.SpreadElement:
                     var newArgument = ReinterpretExpressionAsPattern(expr.As<SpreadElement>().Argument);
-                    node = new RestElement(newArgument.As<Expression>());
+                    node = new RestElement(newArgument.As<IArrayPatternElement>());
                     node.Range = expr.Range;
                     node.Location = _config.Loc ? expr.Location : default;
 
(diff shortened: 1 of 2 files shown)


sebastienros:
FYI I found out how to move the hosting logic to Jint. When I was reading the es6 spec it was described, I had never paid attention. It's related to const/let.

sebastienros:
`([...[...x]]) => {}`

WAT !

lahma:
Merged as was blocking Jint dev.

src/Esprima/JavascriptParser.cs
```diff
@@ -27,7 +27,7 @@ private sealed class Context
             public HashSet<string> LabelSet;
         }
 
-        private struct HoistingScopeLists
+        private class HoistingScopeLists
```
atifaziz: What was the reason for this (dramatic) change? Is it to be able to use `ref var` below?

lahma: Smallest repro below. Changing to struct will turn test case red.

```
[Fact]
public void HoistingScopeShouldWork()
{
    var parser = new JavaScriptParser("function p() {}");
    var program = parser.ParseProgram();
    Assert.NotEmpty(program.HoistingScope.FunctionDeclarations);
}
```

atifaziz: Makes sense but why not add that test?

That said, I think there was another way to solve the problem rather than changing the type of the type. One could pop, modify and push back:

```diff
diff --git a/src/Esprima/JavascriptParser.cs b/src/Esprima/JavascriptParser.cs
index 6ec4931..7731b75 100644
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -27,7 +27,7 @@ namespace Esprima
             public HashSet<string> LabelSet;
         }
 
-        private class HoistingScopeLists
+        private struct HoistingScopeLists
         {
             public ArrayList<FunctionDeclaration> FunctionDeclarations;
             public ArrayList<VariableDeclaration> VariableDeclarations;
@@ -3333,9 +3333,10 @@ namespace Esprima
             _context.AllowYield = previousAllowYield;
 
             var functionDeclaration = Finalize(node, new FunctionDeclaration(id, parameters, body, isGenerator, LeaveHoistingScope(), hasStrictDirective));
-            var lists = _hoistingScopes.Peek();
+            var lists = _hoistingScopes.Pop();
             ref var functionDeclarations = ref lists.FunctionDeclarations;
             functionDeclarations.Add(functionDeclaration);
+            _hoistingScopes.Push(lists);
 
             return functionDeclaration;
         }
@@ -4265,9 +4266,10 @@ namespace Esprima
         {
             if (variableDeclaration.Kind == VariableDeclarationKind.Var)
             {
-                var scopeLists = _hoistingScopes.Peek();
+                var scopeLists = _hoistingScopes.Pop();
                 ref var variableDeclarations = ref scopeLists.VariableDeclarations;
                 variableDeclarations.Add(variableDeclaration);
+                _hoistingScopes.Push(scopeLists);
             }
 
             return variableDeclaration;
```

PR #98, opened 2019-03-02
Revert HoistingScopeLists to be struct

atifaziz:
PR #97 changed `HoistingScopeLists` to be a `class` when it could have been kept as a `struct` by slightly changing how it is used. This PR reverts `HoistingScopeLists` to be `struct` again.

---
PS In the commit message, I have [given credit](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors) to @lahma for the test.


lahma:
Looks good, back to struct usage with verified code coverage, thanks for tweaking!

PR #102, opened 2019-05-26
Report invalid RegEx flags eagerly

lahma:
Relates to https://github.com/sebastienros/jint/issues/571 . Invalid flags will produce null regex which Jint cannot handle, would seem more reasonable for JS parsing to prohibit such definition as Jint has no idea what could be the causes for RegEx instance's Value (string) to be null.


lahma:
@sebastienros can you check if this fix is valid, I didn't see a reason not to report flag problems as parsing exception.

PR #116, opened 2019-10-26
Fix Negative Column in Binary Expression String Concatenation

lahma:
https://github.com/jquery/esprima/commit/22926d5a1317a8e0e027bfa479e3dba00a9abfbd


lahma:
Release to NuGet would make consuming on Jint side.

PR #120, opened 2020-02-07
WIP: Async functions support

codeFather2:
I thought that support for asynchronous functions and expressions would be a good addition to the project. For now tests are failing, because off json dumps format change. I will fix it, for now it's just a WIP PR :)
**TODO** : `await` statements

### DIFF
--- a/src/Esprima/Ast/ArrowFunctionExpression.cs
+++ b/src/Esprima/Ast/ArrowFunctionExpression.cs
@@ -12,12 +12,14 @@ namespace Esprima.Ast
         public bool Expression { get; }
         public HoistingScope HoistingScope { get; }
         public bool Strict { get; }
+        public bool Async { get; }
 
         public ArrowFunctionExpression(
             in NodeList<INode> parameters,
             INode body,
             bool expression,
-            HoistingScope hoistingScope) :
+            HoistingScope hoistingScope,
+            bool async) :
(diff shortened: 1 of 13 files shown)


sebastienros:
Great to see! Are you following the esprima source code?

codeFather2:
I didn’t look at the original repository when I started to do it, but since it turned out to be quite similar to the original, I will finish writing it off from the original. For some reason I didn’t think to do it right this way.

sebastienros:
This is how the whole parser has been written, almost direct port from js to c#. And it makes it easier for us to apply fixes locally. Including the unit tests.

lahma:
Might help to update failing parse trees with ones from JS esprima master, they have the async attribute included IIRC..

lahma:
I went through the Esprima tree and checked required changes, unfortunately their approach was quite different I was unable to refine without a rewrite. I've created a separate PR which shows the changes required to get tests passing and code conforming to original project style https://github.com/sebastienros/esprima-dotnet/pull/121 . 

@codeFather2 I hope you don't mind I went ahead with this with different approach. It can be tricky to port the features as the documentation is a bit lacking about best practices here.

sebastienros:
@codeFather2 thanks for your effort

codeFather2:
@lahma thank you for your work 👍

PR #123, opened 2020-03-04
Support netstandard2.0 and net461 only

lahma:
It's year 2020 and .NET 4.6.1 is the current oldest officially supported full framework. It also has Span support so we shouldn't be tied to the past.

### DIFF
--- a/src/Esprima/Esprima.csproj
+++ b/src/Esprima/Esprima.csproj
@@ -7,7 +7,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <BuildNumber Condition="'$(BuildNumber)' == ''">0</BuildNumber>
     <Authors>Sebastien Ros</Authors>
-    <TargetFrameworks>net45;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


src/Esprima/Esprima.csproj
```diff
@@ -7,7 +7,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <BuildNumber Condition="'$(BuildNumber)' == ''">0</BuildNumber>
     <Authors>Sebastien Ros</Authors>
-    <TargetFrameworks>net45;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
```
sebastienros: Any reasoning behind this choice?
It also means we'll have to change it too in Jint (currently netstandard2.0;net45)

lahma: .NET 4.6.1 has the support for netstandard2.0 (even though a bit flaky at times). Full framework only has support (based on OS support policy) for 4.6 and later. With minimum of 4.6.1 System.Memory becomes accessible and thus, Span.

Generally what I've heard, Microsoft is also advocating ditching netstandard1.x altogether as more of a beta version of the library idea.

I think this would be a good choice for Jint too. Semantic versioning can be troublesome if minor version breaks supported platforms.

PR #124, opened 2020-05-02
Remove hoisting scope handling

lahma:
Removed code related to hoisting. This will be handled by Jint internally and we don't want to pay the price twice.

* rename ExceptionHelper to EsprimaExceptionHelper to allow shared-source scenarios, using static
* cleanup exception handling a bit
* remove obsolete ParseProgram


src/Esprima/Ast/Nodes.cs
```diff
@@ -1,74 +1,83 @@
...
-        ImportNamespaceSpecifier,
-        ImportDeclaration,
-        ExportSpecifier,
-        ExportNamedDeclaration,
-        ExportAllDeclaration,
-        ExportDefaultDeclaration,
-        ClassExpression,
-    };
+        // should not recurse into elements with this flag
+        internal const byte SkipHoistingRecursion = 1 << 7;
```
sebastienros, line 9: Where is it used/

lahma: I have code prepared on Jint's side that checks in AST walker whether should do a recursion into node, when known to be, say, literal or identifier extra call won't be made and call stack doesn't go deeper or child node iterator won't be created. I can't push the code before I have a target Esprima available....

lahma: The ScriptWalker will change into this:

```
        // internal Esprima flag we can use
        private const byte SkipHoistingRecursion = 1 << 7;

        private sealed class ScriptWalker
        {
            internal List<FunctionDeclaration> _functions;

            private readonly bool _collectVarNames;
            internal List<VariableDeclaration> _variableDeclarations;
            internal List<string> _varNames;

            private readonly bool _collectLexicalNames;
            internal List<VariableDeclaration> _lexicalDeclarations;
            internal List<string> _lexicalNames;

            public ScriptWalker(bool collectVarNames, bool collectLexicalNames)
            {
                _collectVarNames = collectVarNames;
                _collectLexicalNames = collectLexicalNames;
            }

            public void Visit(INode node, bool root)
            {
                foreach (var childNode in node.ChildNodes)
                {
                    if (childNode is null)
                    {
                        // array expression can push null nodes in Esprima
                        continue;
                    }

                    if (childNode is VariableDeclaration variableDeclaration)
                    {
                        if (variableDeclaration.Kind == VariableDeclarationKind.Var)
                        {
                            _variableDeclarations ??= new List<VariableDeclaration>();
                            _variableDeclarations.Add(variableDeclaration);
                            if (_collectVarNames)
                            {
                                _varNames ??= new List<string>();
                                ref readonly var nodeList = ref variableDeclaration.Declarations;
                                foreach (var declaration in nodeList)
                                {
                                    if (declaration.Id is Identifier identifier)
                                    {
                                        _varNames.Add(identifier.Name);
                                    }
                                }
                            }
                        }

                        if (root && variableDeclaration.Kind != VariableDeclarationKind.Var)
                        {
                            _lexicalDeclarations ??= new List<VariableDeclaration>();
                            _lexicalDeclarations.Add(variableDeclaration);
                            if (_collectLexicalNames)
                            {
                                _lexicalNames ??= new List<string>();
                                ref readonly var nodeList = ref variableDeclaration.Declarations;
                                foreach (var declaration in nodeList)
                                {
                                    if (declaration.Id is Identifier identifier)
                                    {
                                        _lexicalNames.Add(identifier.Name);
                                    }
                                }
                            }
                        }
                    }
                    else if (childNode is FunctionDeclaration functionDeclaration)
                    {
                        _functions ??= new List<FunctionDeclaration>();
                        _functions.Add(functionDeclaration);
                    }

                    if (((int) childNode.Type & SkipHoistingRecursion) != 0)
                    {
                        continue;
                    }

                    Visit(childNode, false);
                }
            }
        }

```

lahma:
I'm going to do a separate refactoring PR soon that changes node class models a bit.

sebastienros:
I would suggest to target the same 2.0 branch (aka master once release/1.x is created)

sebastienros:
Branches `release/1.x` and `release/2.x` created, in case you'd think that about making `master` the default  development branch.

lahma:
I'm fine with targeting either master or dev for PR.

PR #136, opened 2020-10-21
Adding API documentation

sebastienros:


src/Esprima/JavascriptParser.cs
```diff
@@ -6,6 +6,12 @@
 
 namespace Esprima
 {
+    /// <summary>
+    /// Provides JavaScript parsing capabilities.
+    /// </summary>
+    /// <remarks>
+    /// Use the methods <c>ParseScript()</c>, <c>ParseModule()</c> or <c>ParseExpression()</c> to parse the JavaScript code.
```
lahma, line 13: could also use <see cref="" /> syntax

lahma (approved):

PR #138, opened 2020-10-31
Fix multi-line comment scanning

### LINKED ISSUE #137: System.NullReferenceException when scanning comments in v1.0.1270
The code
 ```csharp
    var scanner = new Scanner("var foo=1; /* \"330413500\" */", new ParserOptions {Comment = true, Loc = true});
    Token token;
    do
    {
        foreach (var comment in scanner.ScanComments())
            Console.WriteLine($"{comment.Start}-{comment.End}");
        token = scanner.Lex();
    } while (token.Type != TokenType.EOF);
 ```

works with Esprima v1.0.1258 (prints `11-28`) and doesn't work with v1.0.1270 where the exception is thrown:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Esprima.Scanner.SkipMultiLineComment() in C:\projects\esprima-dotnet\src\Esprima\Scanner.cs:line 317
   at Esprima.Scanner.ScanCommentsInternal() in C:\projects\esprima-dotnet\src\Esprima\Scanner.cs:line 425
   at Esprima.Scanner.ScanComments() in C:\projects\esprima-dotnet\src\Esprima\Scanner.cs:line 383
   at ConsoleApp1.Program.Main(String[] args) in C:\sandbox\EsprimaCheck\ConsoleApp1\Program.cs:line 15
```

lahma:
Nullable annotations were yelling that this is a bad idea, and it was. 

Original code used to just set line and column but the real problem was that different data structures were used than in original Esprima. Changed to follow the same format for scanner structures as in JS version:

https://github.com/jquery/esprima/blob/45c9ab14d96f7f7fa88333fdd897487a8c20082f/src/scanner.ts#L15-L31

fixes #137

### DIFF
--- a/src/Esprima/Marker.cs
+++ b/src/Esprima/Marker.cs
@@ -1,6 +1,6 @@
 namespace Esprima
 {
-    public class Marker
+    internal class Marker
     {
         public int Index;
         public int Line;
--- a/src/Esprima/Marker.cs
+++ b/src/Esprima/Marker.cs
@@ -1,0 +1,20 @@
+namespace Esprima
+{
+    public class Marker
+    {
+        public int Index;
+        public int Line;
+        public int Column;
+
+        public Marker()
+        {
+        }
+
+        public Marker(int index, int line, int column)
+        {
+            Index = index;
+            Line = line;
+            Column = column;
+        }
+    }
+}
--- a/src/Esprima/Utils/AstJson.cs
+++ b/src/Esprima/Utils/AstJson.cs
@@ -168,30 +168,30 @@ namespace Esprima.Utils
                 {
                     if (includeRange)
                     {
-                        writer.Member("range");
-                        writer.StartArray();
-                        writer.Number(node.Range.Start);
-                        writer.Number(node.Range.End);
-                        writer.EndArray();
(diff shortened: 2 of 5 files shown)


src/Esprima/Utils/AstJson.cs
```diff
@@ -168,30 +168,30 @@ void WriteLocationInfo(Node node)
                 {
                     if (includeRange)
                     {
-                        writer.Member("range");
```
lahma, line 171: compiler issued a warning as the member field is checked against null but the parameter is not (these use parameter event though member has been initialized)

src/Esprima/Marker.cs
```diff
@@ -0,0 +1,20 @@
+namespace Esprima
+{
+    public class Marker
```
atifaziz, line 3: This doesn't need to be public.

atifaziz (changes requested):

src/Esprima/Marker.cs
lahma: Made it internal, good catch

atifaziz (approved):

PR #142, opened 2020-12-13
Support nullish coalescing

lahma:
https://github.com/jquery/esprima/commit/3515f7e596a351349235517d5435e834c8d395b7

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,6 @@ build_script:
   - dotnet --version
   - dotnet pack -c Release
 test_script:
-  - dotnet test .\test\Esprima.Tests\Esprima.Tests.csproj -c Debug
   - dotnet test .\test\Esprima.Tests\Esprima.Tests.csproj -c Release
 artifacts:
   - path: 'src\**\*.*nupkg'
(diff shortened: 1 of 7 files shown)


sebastienros (approved):

sebastienros:
No test?

lahma:
They run parsing against the 262 test suite like jint with allow list to skip unsupported features, NET version doesn't have this.

atifaziz (changes requested):
This PR upgrades projects to target .NET 5 (silently dropping .NET Core 3.1 that is LTS) , which is an unrelated to its subject (“Support nullish coalescing”) and therefore should be a separate PR.

sebastienros:
@atifaziz I don't think it's dropping anything. Just that the tests and benchmarks use the latest version. The library keeps the current targets.

sebastienros:
However to be thorough we might want to run the tests also on 3.1 since there were breaking changes with ICU in 5.0.

lahma:
Restored the netcoreapp3.1 target for tests and removed Debug mode testing in appveyor as we only ship Release mode builds.

atifaziz (approved):
> @atifaziz I don't think it's dropping anything. Just that the tests and benchmarks use the latest version. The library keeps the current targets.

I find the principle of not introducing unrelated changes (even if harmless) keeps reviews clear & focused, but can let it pass if it's acceptable here.

PR #148, opened 2021-02-11
AstVisitor refactoring

MaceWindu:
This PR contains refactored `AstVisitor` class that:
- pass all nodes through `Visit(Node)` dispatcher. Previous implementation were cutting corners a lot making it hard to use without a lot of overloads
- update all `Visit<Node>` methods to enumerate all child nodes. Previous implementation skipped a lot of nodes
- tracks stack of parent nodes. Replaces `NodeExtensions.AncestorNodes` helpers which traverse whole AST on each call

It's kinda breaking one (no issues with existing tests, but public API surface and behavior changed), but I think it is acceptable for v2 release. Other option is to have two visitors: legacy one and new.

### DIFF
--- a/src/Esprima/Utils/AstVisitor.cs
+++ b/src/Esprima/Utils/AstVisitor.cs
@@ -7,28 +7,8 @@ namespace Esprima.Utils
 {
     public class AstVisitor
     {
-        private readonly List<Node> _parentStack = new List<Node>();
-        protected IReadOnlyList<Node> ParentStack => _parentStack;
-
-        /// <summary>
-        /// Returns parent node at specified position.
-        /// </summary>
-        /// <param name="offset">Zero index value returns current node; one corresponds to direct
-        /// parent of current node.</param>
-        protected Node? TryGetParentAt(int offset)
-        {
-            if (_parentStack.Count < offset + 1)
-            {
-                return null;
-            }
-
-            return _parentStack[_parentStack.Count - 1 - offset];
-        }
-
         public virtual void Visit(Node node)
         {
-            _parentStack.Add(node);
-
             switch (node.Type)
             {
                 case Nodes.AssignmentExpression:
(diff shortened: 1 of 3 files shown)


jogibear9988:
@sebastienros 
any change to get this merged? I wanted to work on inclusion of esprimat dotnet into my software. but if this will be merged, I will take this as a basis

src/Esprima/Utils/AstVisitor.cs
```diff
@@ -1,15 +1,34 @@
...
+using System.Runtime.CompilerServices;
 using Esprima.Ast;
-using static Esprima.EsprimaExceptionHelper;
 
 namespace Esprima.Utils
 {
     public class AstVisitor
     {
-        public bool IsStrictMode { get; set; } = false;
+        private readonly List<Node> _parentStack = new List<Node>();
```
sebastienros, line 10: Doesn't seem to be used. Am I missing something?

jogibear9988:
@sebastienros 
Think it's there that you can use it if you derive from it.

src/Esprima/Utils/AstVisitor.cs
MaceWindu: This is used to track parent nodes stack instead of `NodeExtensions.AncestorNodes` call, but I think it is not necessary here as could be implemented easily by deriving from visitor. Should I remove it?

lahma: Maybe if it's unused, private and not referenced might be easier to remove 🙂

MaceWindu: done

lahma:
@MaceWindu https://github.com/sebastienros/esprima-dotnet/pull/176 was merged so could rebase with main/otherwise check if there are still bits missing. I took the approach with some optimizations and proper test coverage so that we can always visit the code/constructs that Esprima supports.

MaceWindu:
I'll close it for now. Will open new PR if there are still things missing when I will have some time, but it looks good.

PR #155, opened 2021-04-29
Support for ES2018 Template Literal Revision

lahma:
* https://github.com/jquery/esprima/commit/ab6420b85cced6b55730a5204594683a17941108
* also fixed curly stack behavior as we actually need a list to achieve both first and last peeking behavior


lahma:
Merged as safe and follows JS version closely, even passes tests.

PR #166, opened 2021-05-19
Fix ParseCatchClause duplicate binding check

lahma:
Was unfortunately broken while merging optional catch clause changes.


lahma:
Fix verified on Jint side, using larger try statement suite.

PR #168, opened 2021-06-14
Fix regular expressions with /[^]/

### LINKED ISSUE #146: ParserException: InvalidRegularExpression on valid regex
Regular expression literal:
```js
/(?:\s|\/\/.*|\/\*[^]*?\*\/)*/g
```

Source `@babel\parser\lib\index.js`
```js
const skipWhiteSpace = /(?:\s|\/\/.*|\/\*[^]*?\*\/)*/g;  
```

[index.zip](https://github.com/sebastienros/esprima-dotnet/files/5966005/index.zip)

sebastienros:
Fixes #146

### DIFF
--- a/src/Esprima/Scanner.cs
+++ b/src/Esprima/Scanner.cs
@@ -1549,6 +1549,13 @@ namespace Esprima
                 tmp = Regex.Replace(tmp, "[\uD800-\uDBFF][\uDC00-\uDFFF]", astralSubstitute);
             }
 
+            // .NET doesn't support [^] which is equivalent to `[^.]`
+            // c.f. https://github.com/sebastienros/esprima-dotnet/issues/146
+            if (tmp.Contains("[^]"))
+            {
+                tmp = tmp.Replace("[^]", "[^.]");
+            }
+
             // First, detect invalid regular expressions.
             var options = ParseRegexOptions(flags);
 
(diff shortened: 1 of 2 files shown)


src/Esprima/Scanner.cs
```diff
@@ -1549,6 +1549,13 @@ public Token ScanTemplate()
                 tmp = Regex.Replace(tmp, "[\uD800-\uDBFF][\uDC00-\uDFFF]", astralSubstitute);
             }
 
+            // .NET doesn't support [^] which is equivalent to `[^.]`
+            // c.f. https://github.com/sebastienros/esprima-dotnet/issues/146
+            if (tmp.Contains("[^]"))
```
lahma, line 1554: not sure if calling Replace directly is much different, search must be done again

lahma (approved):

src/Esprima/Scanner.cs
sebastienros: I just checked and it looks like it's allocating in all cases. https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs#L1042

It's probably insignificant, but I think we can keep the test then.

lahma: Sure, I was just wondering.

Issue #169, opened 2021-06-20
fork of esprima


jogibear9988:
@sebastienros 
as the development on esprima is very slow, I'll created a fork "esprima-next" where i merged some of the pull's and started to work on some features (class properties/private properties https://github.com/node-projects/esprima-next/pulls)
maybe this is also interesting for you

jogibear9988:
if now created pull reqs here for the functional changes

PR #170, opened 2021-06-27
Fix parsing of static member expression property initializer

### LINKED ISSUE #145: ArgumentOutOfRangeException parsing class member declared using Symbol
```
Exception of type 'System.ArgumentOutOfRangeException' was thrown.
Parameter name: Type
Actual value was MemberExpression.
```

Exception generated when I tried to parse node.js `node_modules` folder (to test parser-dependent application).
Original file: `node_modules\npm\node_modules\@npmcli\arborist\lib\edge.js` from node.js distribution

[edge.zip](https://github.com/sebastienros/esprima-dotnet/files/5965824/edge.zip)

Offending lines (looks like failure on method name parse):
```js
  [util.inspect.custom] () {
    return this.toJSON()
  }
```

lahma:
The original implementation handles gracefully of not having expected type, as does the following logic of the method. There's no need to throw exception if we cannot extract the id.

fixes #145


jogibear9988:
@lahma 
only for your info, i created a fork of esprima, cause development seemed go have slowed down...
https://github.com/sebastienros/esprima-dotnet/issues/169

lahma:
@jogibear9988 yes I noticed that. It's sad that it takes such a long time for the JS version to handle PRs, but it's OSS so we get what we pay for. Separate fork might get hard to synchronize but I hope it serves the purpose you need with its features.

lahma:
Verified that fixes also test suites in test262 on Jint side.

PR #172, opened 2021-07-11
Fix adapting Regex for C#

KurtGokhan:
Jint was still suffering from the `[^]` bug. The recent fix by @sebastienros  was preventing Parser to fail but it was not returning the adapted Regex.

This code seems confusing to me. Why do we parse the Regex twice? It is the same in original `esprima`.

Can't we just parse the Regex in one pass?

### DIFF
--- a/src/Esprima/Scanner.cs
+++ b/src/Esprima/Scanner.cs
@@ -1574,20 +1574,29 @@ namespace Esprima
             try
             {
                 // Do we need to convert the expression to its .NET equivalent?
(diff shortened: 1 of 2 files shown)


lahma (approved):
Looking good, I don't think there's a good reason to parse twice. Code mostly follows the JS version for maintainability and there can be some quirks.

KurtGokhan:
@lahma  can you release this to nuget? I am planning to make related changes to Jint.

lahma:
@KurtGokhan there's a new 2.0.1 release now on NuGet

PR #173, opened 2021-07-13
Fix [^] and [] patterns in Regex

KurtGokhan:
I noticed some mistakes in my previous PR when testing Jint.

- `[^]` pattern (which should match all characters) was converted to `[^.]` (which does not match `.` character)
- `[]` pattern was still not handled

Note: `[^]` pattern normally works in .NET 5 but not .NET 4.


KurtGokhan:
Actually, let's not merge this for now. I noticed that blindly replacing the regex can cause valid regex to be altered. For example:

- `/[[^]/` is valid but it will turn into `/[[\s\S]/`
- `/[[]/` is valid but it will turn into invalid `/[(?:a^)/`

Seems like this is an impossible task without something like a regex state machine.

lahma:
I've marked this as draft so that accidental merge doesn't happen until ready. Thanks for investigating the issue!

KurtGokhan:
I reverted some changes from my previous PR. I also added some additional checks.

Esprima now will not fail parsing the entire program when parsing these two patterns. But it will return `null` as the parsed regex to the caller, e.g. Jint. 

I opened a PR in Jint to handle null Regex differently. sebastienros/jint#931

lahma (approved):

PR #175, opened 2021-07-22
Allow reusing same JavaScriptParser instance for parsing

lahma:
It's always been a bit odd that you need to instantiate a new instance of `JavaScriptParser` every time you want to parse something. Changing parsing to be stateless and based on `ParseScript`/`ParseModule` input, resetting internal state between calls. Tested on Jint side and all tests green when reusing same parser instance inside engine instance.

API breaking change so probably requires setting version to 3.x.

Before (new'ing parser 200 times for parsing same script):

|       Method |   N |     Mean |     Error |    StdDev |    Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------- |---- |---------:|----------:|----------:|---------:|-------:|------:|----------:|
| ParseProgram | 200 | 2.639 ms | 0.0051 ms | 0.0043 ms | 312.5000 | 3.9063 |     - |   5.03 MB |

After (having parser as member and reusing same instance):

|       Method |   N |     Mean |     Error |    StdDev |    Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------- |---- |---------:|----------:|----------:|---------:|-------:|------:|----------:|
| ParseProgram | 200 | 2.420 ms | 0.0051 ms | 0.0048 ms | 292.9688 | 3.9063 |     - |   4.68 MB |

### DIFF
--- a/samples/Esprima.Benchmark/AstTreeWalkBenchmark.cs
+++ b/samples/Esprima.Benchmark/AstTreeWalkBenchmark.cs
@@ -14,8 +14,8 @@ namespace Esprima.Benchmark
         public void Setup()
         {
             var code = File.ReadAllText("3rdparty/angular-1.7.9.js");
-            var parser = new JavaScriptParser(code);
-            _script = parser.ParseScript();
+            var parser = new JavaScriptParser();
+            _script = parser.ParseScript(code);
         }
 
         [Benchmark]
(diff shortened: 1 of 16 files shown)


adams85:
As I mentioned in the issue above, I believe this would be a nice improvement. I'm just curious why you decided to cancel it despite the cleaner API and performance gains. It's too late for such a BC now maybe?

Anyway, it would be nice to have this in v3, so I'm willing to resurrect this if you're ok with it. But, of course, I don't want to push it, so I'll let it go if you still think it's not a good idea for some reason.

lahma:
I think it would be a valid addition. I closed because gains were a bit marginal and again some merge conflicts, but no reason not to resurrect.

PR #176, opened 2021-07-23
Add missing branching to AstVisitor

lahma:
While investigating function code analysis for Jint and skipping creation of arguments instance based on AST analysis, found out that the AstVisitor was quite lacking and throwing errors. Added test to visit all test JS files which showed multiple problems. Now both Esprima and Jint are green for AST analysis for code base.

* Add missing branches
* Remove `VisitStatement` and `VisitExpression`, confusing like noted and can be worked around by overriding `Visit`
* Optimize looping code

Will conflict with #148 but I think we can refresh that after this and see if there's more to be done.

### DIFF
--- a/src/Esprima/Ast/ArrayPattern.cs
+++ b/src/Esprima/Ast/ArrayPattern.cs
@@ -2,15 +2,17 @@
 {
     public sealed class ArrayPattern : BindingPattern
     {
-        private readonly NodeList<Expression> _elements;
+        private readonly NodeList<Expression?> _elements;
 
-        public ArrayPattern(in NodeList<Expression> elements) : base(Nodes.ArrayPattern)
+        public ArrayPattern(in NodeList<Expression?> elements) : base(Nodes.ArrayPattern)
         {
             _elements = elements;
         }
 
-        public ref readonly NodeList<Expression> Elements => ref _elements;
+        public ref readonly NodeList<Expression?> Elements => ref _elements;
 
+#pragma warning disable 8631
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(_elements);
+#pragma warning restore 8631
     }
 }
--- a/src/Esprima/Utils/AstJson.cs
+++ b/src/Esprima/Utils/AstJson.cs
@@ -585,7 +585,7 @@ namespace Esprima.Utils
                 }
             }
 
-            protected override void VisitFunctionExpression(FunctionExpression function)
+            protected override void VisitFunctionExpression(IFunction function)
             {
                 using (StartNodeObject((Node) function))
                 {
--- a/src/Esprima/Utils/AstJson.cs
+++ b/src/Esprima/Utils/AstJson.cs
@@ -249,10 +249,10 @@ namespace Esprima.Utils
                 Member(name, map[value]);
             }
 
-            private void Member<T>(string name, in NodeList<T> nodes) where T : Node =>
+            private void Member<T>(string name, in NodeList<T> nodes) where T : Node? =>
                 Member(name, nodes, node => node);
 
-            private void Member<T>(string name, in NodeList<T> list, Func<T, Node> nodeSelector) where T : Node
+            private void Member<T>(string name, in NodeList<T> list, Func<T, Node?> nodeSelector) where T : Node?
             {
                 Member(name);
                 _writer.StartArray();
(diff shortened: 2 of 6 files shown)


adams85:
You may be interested in this: https://gist.github.com/adams85/a1dead2105788d0ea7204326fafa5ec1

I created this based on your changes but did some further fixes and improvements.

lahma:
> @lahma You may be interested in this: https://gist.github.com/adams85/a1dead2105788d0ea7204326fafa5ec1

Thanks for the pointer. I hopefully have integrated most of your pointers, except for arrow expression handling

lahma:
@sebastienros please take a look.

adams85:
@lahma Although I'm not @sebastienros, I'd like to add my remarks.

First of all, I advise against shortcuts like [this one](https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L246) in `AstVisitor`. Even when the concrete `VisitXXX` method is known based on the type of the node to visit, I suggest routing every visitation through the `Visit` method because this would enable subclasses to add generic behavior (e.g. tracking parent nodes) by overriding `Visit` only.

Further minor issues I found:

* `VisitFunctionDeclaration` should visit `Id` (if it's not null).

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L259

* `VisitLabeledStatement` should visit `Label`.

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L342

* Arrow functions can enable strict mode just like regular functions, so strict mode should be determined by the parser, not by the visitor, which can know about that in no way beforehand. (I mean the visitor's IsStrictMode property seems to make no sense.) See also https://github.com/sebastienros/esprima-dotnet/issues/179

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L418

* Arrow functions can be async.

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L419

* `VisitFunctionExpression` should visit `Id` (if it's not null).

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L495

* `VisitClassExpression` should visit `SuperClass` (if it's not null) and `Body`.

https://github.com/sebastienros/esprima-dotnet/blob/ed4f4b954d92f3c0153e0336b574f764fba979c9/src/Esprima/Utils/AstVisitor.cs#L512

lahma:
@adams85 thanks for the detailed feedback! I've incorporated you suggestions in the latest commit, again skipping the strict mode handling as I think it's a separate issue due to requiring parser code changes.

adams85:
@lahma It all looks good except for https://github.com/sebastienros/esprima-dotnet/pull/176/commits/2b61fc9e0f82e756f587148c34b3254276b880d1#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR419

Here we don't actually visit a child node in the AST but a substitute for the original node. So I think it should stay as it was or the code from `VisitFunctionExpression` should be copied over. Maybe the latter is the better solution as arrow functions don't ever have a name identifier so that can be omitted safely.

One more remark:  The change https://github.com/sebastienros/esprima-dotnet/pull/176/commits/2b61fc9e0f82e756f587148c34b3254276b880d1#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR489 breaks existing code. I'm not sure it's worth it.

lahma:
@adams85 thanks, I pushed another tweak, I reverted back to having `VisitFunctionExpression(IFunction function)`  even though II felt it is and anomality because other `IFunction` visitors (`ArrowFunctionExpression` and `FunctionDeclaration`) visit the concrete type and this one the interface even though only invoked with a concrete type.

adams85:
@lahma Yeah, `VisitFunctionExpression(IFunction function)` looks a bit weird but the method name makes it unambigous after all.

lahma:
@adams85 if you don't find anything blocking, I think we can probably soon merge as there has been a peer review after all.

adams85:
@lahma I took another look. Seems fine to me.

Issue #40, opened 2018-09-23
Comments are not handled at all?


KvanTTT:
Consider the following javascript code:

```csharp
// c
```

I'm trying to get all comments with the following code:

```csharp
var parserOptions = new ParserOptions(source) { Comment = true, Tokens = true };
var scanner = new Scanner(source, parserOptions);
var comments = scanner.ScanComments();
```

But getting an empty `comments` collection.

The following code also does not work (from Esprima.Sample):

```csharp
do
{
    scanner.ScanComments();
    token = scanner.Lex();
    tokens.Add(token);
} while (token.Type != TokenType.EOF);
```

Returns only a single empty token:

```json
[
  {
    "Type": 1,
    "Start": 4,
    "End": 4,
    "LineNumber": 1,
    "Location": {}
  }
]
```

How can I handle them?

KvanTTT:
Finally, I found a way to extract all comments:

```csharp
var comments = new List<Comment>();
do
{
    comments.AddRange(scanner.ScanComments());
    token = scanner.Lex();
    tokens.Add(token);
} while (token.Type != TokenType.EOF);
```

But:

* All comment values `Comment.Value` are `null`, but should be filled.
* Lexer goes into infinity loop on some files with custom `ErrorHandler` that does not throw an exception on the first error. For example, [angular-1.2.5.js](https://github.com/sebastienros/esprima-dotnet/blob/dev/test/Esprima.Tests/Fixtures/3rdparty/angular-1.2.5.js) file from this repository.

sebastienros:
I think the issue is related to this PR that I seem to have skipped https://github.com/jquery/esprima/pull/1439/files

Each node is passed to `proxyDelegate` such that external components can react to each parsed node. The a composite one will also plug `commentHandler` to it to attach comments to their statement/expression. As a general rule I think we should have that in place. It might make hoisting easier to implement.

adams85:
Checked and the following code works just as expected:

```js
var scanner = new Scanner(source, new ScannerOptions { Comments = true });
var comments = scanner.ScanComments();
```

For extracting all the comments, you can do what you invented [here](https://github.com/sebastienros/esprima-dotnet/issues/40#issuecomment-423843683) but please be aware that tokenization is not always possible. (For more details, see https://github.com/sebastienros/esprima-dotnet/issues/44#issuecomment-1445116490)

A reliable way to extract comments is the following:

```js
[...]

Issue #39, opened 2018-09-23
Attach leading and trailing comments to AST nodes


KvanTTT:
### Input

Parse the following code:

```javascript
// leading
a // trailing
```

with turned on `Comment = true` option in `ParserOptions`:

```csharp
var program = parser.ParseProgram(source, new ParserOptions(source) { Comment = true; });
```

After that, serialize the result AST with `JsonConvert`.

### Expected

Leading and trailing should be attached to AST nodes. Check it on [Esprima.org](http://esprima.org/demo/parse.html?code=%2F%2F%20leading%0Aa%20%2F%2F%20trailing).

```json
{
    "type": "Program",
    "body": [
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Identifier",
                "name": "a",
                "trailingComments": [
                    {
                        "type": "Line",
                        "value": " trailing",
                        "range": [
                            13,
                            24
                        ]
                    }
                ]
            },
            "leadingComments": [
                {
                    "type": "Line",
                    "value": " leading",
                    "range": [
                        0,
                        10
                    ]
                }
            ]
        }
    ],
    "sourceType": "script"
}
```

### Actual

Comments are completely skipped.

```json
{
  "Type": "Program",
  "Body": [
    {
      "Type": "ExpressionStatement",
      "Expression": {
        "Type": "Identifier",
        "Name": "a",
        "Loc": {}
      },
      "Loc": {}
    }
  ],
  "SourceType": "script",
  "Loc": {}
}
```

adams85:
A few thoughts for those who would make an attempt on implementing this: now there's everything in place to port [`CommentHandler`](https://github.com/node-projects/esprima-next/blob/818367b5f291e30dd0441d5f8f80bad306e23cd7/src/comment-handler.ts) from Esprima JS. But be aware that it'll be capable of an approximation at best. There's no way to precisely represent comments in the current AST model because it's simply not detailed enough: adding `LeadingComments`/`TrailingComments` to AST nodes is not enough to describe pieces of syntax like: `async /*some comment*/ function f() { ... }`. To handle such cases, AST nodes should also include tokens as child nodes. (I suggest checking out how the problem is solved in Roslyn.)

However, it's unlikely that proper comment handling will ever be possible in this lib because its main consumer is Jint, which wouldn't want to pay the price of a more detailed AST model. The best we can have is `CommentHandler` but it will be a half-assed solution. Which, of course, still could be useful in some use cases.

[...]

PR #178, opened 2021-07-29
JavaScriptParser collects comments if options.Comments is true.

rohan-warang:
I have use case where the comments in a script are important. As esprima has an [option](https://docs.esprima.org/en/latest/syntactic-analysis.html?highlight=comments#comment-collection) to return an array of comments, I tried to use the `options.Comments` for `JavaScriptParser`. However it seems that the `JavaScriptParser` scans comments and discards them even if the options is set to true. I am suggesting adding a new property JavaScriptParser.Comments for storing and exposing them as a list instead of discarding them.

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -64,6 +64,16 @@ public partial class JavaScriptParser
     /// </remarks>
     public IReadOnlyList<Token> Tokens => _tokens;
 
+    private protected readonly List<Comment> _comments = new();
+
+    /// <summary>
+    /// Returns the list of comments that were parsed.
+    /// </summary>
+    /// <remarks>
+    /// It requires the parser options to be configured to generate comments.
+    /// </remarks>
+    public IReadOnlyList<Comment> Comments => _comments;
+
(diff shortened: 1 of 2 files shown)


lahma:
Thanks for looking into this and creating the PR. This is a bit of an oddball as the [CommentHandler](https://github.com/jquery/esprima/blob/main/src/comment-handler.ts) from Esprima hasn't been ported and there are reports of missing functionality: https://github.com/sebastienros/esprima-dotnet/issues/40 , https://github.com/sebastienros/esprima-dotnet/issues/39 .

What are your thoughts, should the implementation be closer to JS version and handle a bit more specific cases too like it seems to be in source code, I haven't dug into this at all so I have no idea, @sebastienros might have a say on this.

rohan-warang:
I suppose you are looking at supporting the `attachComment` option in Esprima. That would be a good feature if enough people are interested. 
For this PR I think the main concern for me was handling the `comment` option. Since the parser already does all the work required to find the comment node, we might as well support this option by returning the list of collected comments.

adams85:
Rebased and merging this as part of https://github.com/sebastienros/esprima-dotnet/issues/300. Further cleanup will follow in subsequent PRs.

PR #181, opened 2021-08-13
Fix strict mode detection for arrow functions and object methods

### LINKED ISSUE #179: Bug in strict mode tracking?
If I'm not missing something, there's a minor bug related to strict mode tracking: in the case of arrow functions `Strict` property is never set, thus, always return false. See https://github.com/sebastienros/esprima-dotnet/blob/75e35172b20b95fdaf8c94fa5366d11b61a9c249/src/Esprima/Ast/ArrowFunctionExpression.cs#L7

lahma:
Test coverage for missing strict mode detection and fixes to handle.

fixes #179

/cc @adams85

### DIFF
--- a/src/Esprima/Ast/ArrowFunctionExpression.cs
+++ b/src/Esprima/Ast/ArrowFunctionExpression.cs
@@ -8,6 +8,7 @@
             in NodeList<Expression> parameters,
             Node body,
             bool expression,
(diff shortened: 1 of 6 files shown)


lahma:
Tests are covering, verified on Jint side, JS version does not have the property so no implementation to align with.

PR #182, opened 2021-08-13
Fix visit order for ForStatement and ForOfStatement

### LINKED ISSUE #153: AST Visitor - wrong order?
In the Visit of the For Statement, the Update is Visited before the Body.

Is this intended?

I need to change that if I want to use the Visitor to write back Javascript.

lahma:
fixes #153

@jogibear9988 does this look like what you would expect now?

### DIFF
--- a/src/Esprima/Utils/AstVisitor.cs
+++ b/src/Esprima/Utils/AstVisitor.cs
@@ -367,26 +367,19 @@ namespace Esprima.Utils
 
         protected virtual void VisitForStatement(ForStatement forStatement)
         {
-            if (forStatement.Init != null)
+            if (forStatement.Init is not null)
             {
-                if (forStatement.Init.Type == Nodes.VariableDeclaration)
-                {
-                    Visit(forStatement.Init.As<Statement>());
-                }
-                else
-                {
-                    Visit(forStatement.Init.As<Expression>());
-                }
+                Visit(forStatement.Init);
             }
-            if (forStatement.Test != null)
+            if (forStatement.Test is not null)
             {
                 Visit(forStatement.Test);
             }
-            Visit(forStatement.Body);
-            if (forStatement.Update != null)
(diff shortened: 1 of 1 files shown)


adams85:
@jogibear9988 Is this all you need for writing back Js code? I mean there are still a bunch of nodes which are visited in other order than the order in which they appear in the code (document order). For example, template literals:

https://github.com/sebastienros/esprima-dotnet/blob/ce65e31f8d7d4e7d8b52a7e5817c228b0924ab32/src/Esprima/Utils/AstVisitor.cs#L688-L701

Document order would be `Quasis[0]`, `Expressions[0]`, `Quasis[1]`, `Expressions[1]`, ... (`Quasis` always have one more element than `Expressions`.)

@lahma What do you think about this? Wouldn't it make sense to have consistent document order in the base visitor?

lahma:
I don't have strong opinions, but consistent is always good and I was aiming to close some open issues. I only need visitor to visit everything to get enough info on Jint side, order does not matter.

adams85:
@lahma Fair enough. I don't want to push anything, just seemed to be a good idea once the visitor's getting an overhaul anyway. 

So let's return to the question once the ongoing work has been merged, then I may submit a PR with further refinements if you find any of those useful.

jogibear9988:
I still want to work on this, but at the moment, I've no time...
As you maybe have seen, I also forked esprima, i do a webcomponent designer, ...
hopefully I get back to this in december.

jogibear9988:
And if I work on it, maybe I need to change the order of some visitors. But I think mostly I need to oberride all of them, so the initial order does not matter. But If I need to change, I then maybe also create a pull to change the original one

lahma:
Merging as contains cleanup for `VisitForStatement`, we can open separate issue and/or PR if further refinements are actually needed.

PR #185, opened 2021-08-16
Further improvements to AstVisitor - double dispatch

adams85:
This is a proposal to use the standard double dispatch approach of the visitor design pattern instead of the current switch dispatch.

Performance-wise the two approaches are practically the same:

``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19041.1165 (2004/May2020Update/20H1)
AMD Ryzen 7 2700X, 1 CPU, 16 logical and 8 physical cores
.NET SDK=5.0.400
  [Host]             : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
  .NET 5.0           : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
  .NET Core 3.1      : .NET Core 3.1.18 (CoreCLR 4.700.21.35901, CoreFX 4.700.21.36305), X64 RyuJIT
  .NET Framework 4.8 : .NET Framework 4.8 (4.8.4400.0), X64 RyuJIT


```
|         Method |                Job |            Runtime |     Mean |   Error |  StdDev | Rank |
|--------------- |------------------- |------------------- |---------:|--------:|--------:|-----:|
| SwitchDispatch |           .NET 5.0 |           .NET 5.0 | 562.6 ms | 3.67 ms | 2.86 ms |    1 |
| DoubleDispatch |           .NET 5.0 |           .NET 5.0 | 596.6 ms | 2.22 ms | 2.07 ms |    2 |
| SwitchDispatch |      .NET Core 3.1 |      .NET Core 3.1 | 555.8 ms | 2.46 ms | 2.18 ms |    1 |
| DoubleDispatch |      .NET Core 3.1 |      .NET Core 3.1 | 609.0 ms | 4.74 ms | 4.20 ms |    3 |
| SwitchDispatch | .NET Framework 4.8 | .NET Framework 4.8 | 656.8 ms | 2.55 ms | 2.26 ms |    4 |
| DoubleDispatch | .NET Framework 4.8 | .NET Framework 4.8 | 650.7 ms | 7.25 ms | 6.78 ms |    4 |

(The table shows the time needed for entirely traversing a script of ~60000 nodes one thousand times. So a single visitation took 550-600 µs on average.)

However, from a maintenance point of view, double dispatch has some benefits over switch dispatch:
* it's exhaustive by nature (no need for `VisitUnknownNode`),
* makes impossible to forget about the visitor aspect when adding a new AST node type (because of the abstract `Accept` method of the `Node` base class)

### DIFF
--- a/src/Esprima/Ast/ArrayExpression.cs
+++ b/src/Esprima/Ast/ArrayExpression.cs
@@ -15,6 +15,6 @@ namespace Esprima.Ast
 
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(_elements);
 
-        public override void Accept(AstVisitor visitor) => visitor.VisitArrayExpression(this);
+        protected internal override void Accept(AstVisitor visitor) => visitor.VisitArrayExpression(this);
     }
 }
--- a/src/Esprima/Ast/ArrayExpression.cs
+++ b/src/Esprima/Ast/ArrayExpression.cs
@@ -1,3 +1,5 @@
+using Esprima.Utils;
+
 namespace Esprima.Ast
 {
     public sealed class ArrayExpression : Expression
@@ -12,5 +14,7 @@ namespace Esprima.Ast
         public ref readonly NodeList<Expression> Elements => ref _elements;
 
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(_elements);
+
+        public override void Accept(AstVisitor visitor) => visitor.VisitArrayExpression(this);
     }
 }
--- a/src/Esprima/Ast/ArrayPattern.cs
+++ b/src/Esprima/Ast/ArrayPattern.cs
@@ -1,4 +1,6 @@
-﻿namespace Esprima.Ast
+﻿using Esprima.Utils;
+
+namespace Esprima.Ast
 {
     public sealed class ArrayPattern : BindingPattern
     {
@@ -14,5 +16,7 @@
 #pragma warning disable 8631
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(_elements);
 #pragma warning restore 8631
+
+        public override void Accept(AstVisitor visitor) => visitor.VisitArrayPattern(this);
     }
 }
--- a/src/Esprima/Ast/ArrayPattern.cs
+++ b/src/Esprima/Ast/ArrayPattern.cs
@@ -17,6 +17,6 @@ namespace Esprima.Ast
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(_elements);
 #pragma warning restore 8631
 
-        public override void Accept(AstVisitor visitor) => visitor.VisitArrayPattern(this);
+        protected internal override void Accept(AstVisitor visitor) => visitor.VisitArrayPattern(this);
     }
 }
--- a/src/Esprima/Ast/ArrowFunctionExpression.cs
+++ b/src/Esprima/Ast/ArrowFunctionExpression.cs
@@ -1,4 +1,6 @@
-﻿namespace Esprima.Ast
+﻿using Esprima.Utils;
+
+namespace Esprima.Ast
 {
     public sealed class ArrowFunctionExpression : Expression, IFunction
     {
@@ -31,5 +33,7 @@
         public ref readonly NodeList<Expression> Params => ref _params;
 
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(Params, Body);
+
+        public override void Accept(AstVisitor visitor) => visitor.VisitArrowFunctionExpression(this);
     }
 }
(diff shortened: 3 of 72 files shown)


adams85:
@lahma Thanks for bringing attention to this detail which I completely overlooked.

I re-run the tests with sealed visitors:

``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19041.1165 (2004/May2020Update/20H1)
AMD Ryzen 7 2700X, 1 CPU, 16 logical and 8 physical cores
.NET SDK=5.0.400
  [Host]             : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
  .NET 5.0           : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
  .NET Core 3.1      : .NET Core 3.1.18 (CoreCLR 4.700.21.35901, CoreFX 4.700.21.36305), X64 RyuJIT
  .NET Framework 4.8 : .NET Framework 4.8 (4.8.4400.0), X64 RyuJIT


```
|         Method |                Job |            Runtime |     Mean |   Error |  StdDev | Rank |
|--------------- |------------------- |------------------- |---------:|--------:|--------:|-----:|
| SwitchDispatch |           .NET 5.0 |           .NET 5.0 | 561.8 ms | 3.70 ms | 3.46 ms |    2 |
| DoubleDispatch |           .NET 5.0 |           .NET 5.0 | 550.9 ms | 2.06 ms | 1.72 ms |    1 |
| SwitchDispatch |      .NET Core 3.1 |      .NET Core 3.1 | 545.7 ms | 2.71 ms | 2.40 ms |    1 |
| DoubleDispatch |      .NET Core 3.1 |      .NET Core 3.1 | 551.4 ms | 2.58 ms | 2.29 ms |    1 |
| SwitchDispatch | .NET Framework 4.8 | .NET Framework 4.8 | 654.9 ms | 8.22 ms | 7.69 ms |    4 |
| DoubleDispatch | .NET Framework 4.8 | .NET Framework 4.8 | 587.4 ms | 3.11 ms | 2.91 ms |    3 |

It seems that sealing the classes didn't make a significant impact. Interestingly, double dispatch got a bit faster which contradicts your expectations.

I uploaded the relevant files of the benchmark [here](https://gist.github.com/adams85/6dec6b3eb74fcd3897b319f15b487a96) if you want to take a closer look.

lahma:
So there should be no performance issue here by changing the approach. Would it make sense to make the `Accept` method internal for now so it would hide the double-dispatch implementation detail for now?

adams85:
@lahma 

> Would it make sense to make the Accept method internal for now so it would hide the double-dispatch implementation detail for now?

I think it would. At least I can't imagine a use case which needs to directly call `Accept`. Concrete visitors usually have internal state and in such cases starting a visitation through `Accept` may lead to undefined/erroneous behavior. [`Accept` of BCL's `Expression` being declared as `protected internal`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.accept?view=net-5.0) also shows that this reasoning is probably correct.

However, this also applies more or less to the `Visit` method of the visitor base class as well. I think that one should be `protected internal` as well since concrete visitors often have to introduce an alternative entry point due to the necessity of some initialization/finalization steps. In such cases calling `Visit` directly also may lead to undefined/erroneous behavior. Curiously enough, [`Visit` is usually declared as `public`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expressionvisitor?view=net-5.0).

So let's make then the `Accept` to `protected internal`. It'd be nice to change `Visit` as well but that'd be a BC so let's leave that one alone. I'll update the PR soon according to this.

What gave me some more food for thought is to remove (mark as obsolete for now) `VisitUnknownNode` or not. Can that be a valid use case ever to create some extension node types outside of the parser library? If so, that node type must dispatch the call somewhere in its `Accept` method. What's your opinion on this?

lahma:
Great, it seems that we agree on the internal thing.

> What gave me some more food for thought is to remove (mark as obsolete for now) `VisitUnknownNode` or not. Can that be a valid use case ever to create some extension node types outside of the parser library? If so, that node type must dispatch the call somewhere in its `Accept` method. What's your opinion on this?

I can't easily imagine an imaginary node type that isn't even handled by Esprima parser. Of course easiest to find a use case by breaking things... AST visiting is probably less used in the sense of binary compatibility so I guess we can experiment a bit more here. I'd go for obsolete for now. I think people might have earlier overridden it to bypass the earlier limitations of visiting being broken.

adams85:
@lahma So let's make then the Accept method protected internal. [...] I'll update the PR soon according to this.

Done.

lahma (approved):
Looking good, great to see this AST visiting implementation getting incrementally better!

PR #186, opened 2021-08-16
Further improvements to AstVisitor - document order visitation

adams85:
This proposal makes the visitation order of the child nodes predictible by consistently visiting them as they appear in the code.

So this is rather an aesthetic one but there may be some use cases which benefits from it.


lahma (approved):

lahma:
Thank you!

Issue #180, opened 2021-08-02
AST improvement suggestions


adams85:
I'd welcome the following improvements to the AST:

* Introduction of an `IClass` interface for `ClassDeclaration` and `ClassExpression` for the same purpose which `IFunction` serves considering `FunctionDeclaration` and `FunctionExpression`: for allowing polymorphic handling of these types.

* Moving the `Local` property of `ImportDefaultSpecifier`, `ImportNamespaceSpecifier` and `ImportSpecifier` to their base class `ImportDeclarationSpecifier` to avoid [this](https://github.com/adams85/bundling/blob/d9a649cbeeabcff3810c439971e732ffdddd9bcc/source/Bundling.EcmaScript/Internal/Helpers/VariableScopeBuilder.cs#L181). If this is not acceptable (e.g. because we expect that other types of declaration specifiers may be introduced in the future), introducing an interface can also be a viable alternative.

Are you open for a PR addressing these suggestions?

lahma:
FYI, AST related work got published to NuGet yesterday.

adams85:
Great news! Thank you, Marko!

lahma:
Thank *you* for all the help and iterating towards perfection 🙂

PR #188, opened 2021-08-19
AST polymorphism improvements

adams85:
This one addresses https://github.com/sebastienros/esprima-dotnet/issues/180

Note: I think it's better not to move `Local` identifier of the `Import*Specifier` classes to the base class because unfortunatly it's declared as field and because of that we'd have an ABI BC.

### DIFF
--- a/src/Esprima/Ast/ClassDeclaration.cs
+++ b/src/Esprima/Ast/ClassDeclaration.cs
@@ -2,11 +2,16 @@
 
 namespace Esprima.Ast
 {
-    public sealed class ClassDeclaration : Declaration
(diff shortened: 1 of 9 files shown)


lahma (approved):

lahma:
LGTM, I can merge if ready on your part. I think we can revisit field vs property and the performance (JIT etc) at some point, probably vNext.

adams85:
@lahma That's all for me. Thank you.

> I think we can revisit field vs property and the performance (JIT etc) at some point, probably vNext.

Yeah, at least it'd be nice to make everything either property or field because now it's pretty mixed. :)

lahma:
Thank you once again!

Issue #183, opened 2021-08-15
Parser error on Asynchronous iteration and Asynchronous generators


adams85:
The parser has issues with the aforementioned [ES2018 features](https://exploringjs.com/impatient-js/ch_new-javascript-features.html#new-in-es2018):

* ~~*Asynchronous iteration* works only when `ParserOptions.Tolerant` is set to false, otherwise it results in `ParserException`. Sample input:~~
   ```js
   for await (const x of syncToAsyncIterable(['a', 'b'])) {
     console.log(x);
   }
   ```
  results in error as expected but is parsed fine in async contexts:
  ```js
   async function f() {
     for await (const x of syncToAsyncIterable(['a', 'b'])) {
       console.log(x);
     }
   }
  ```

* *Asynchronous generators* results in in `ParserException` regardless the `ParserOptions.Tolerant` setting. Sample input: 
   ```js
   async function* asyncGen() {
     yield someValue;
   }
   ```

Library version: 2.0.2

lahma:
I think this is a problem also in the original Esprima JS project so maybe need to file an issue there.

adams85:
@lahma You mean it has to be fixed in the original Esprima first? That'd be a bummer because this is kinda blocking for my lib...

BTW, have you seen [this comment](https://github.com/sebastienros/esprima-dotnet/issues/179#issuecomment-898889925)? Could you re-open that issue? (Unfortunately I can't.)

lahma:
> @lahma You mean it has to be fixed in the original Esprima first? That'd be a bummer because this is kinda blocking for my lib...

Can be fixed on either side first, just a good practice to get the fix ported also to the other side to keep code consistent between the implementations.

> BTW, have you seen [this comment](https://github.com/sebastienros/esprima-dotnet/issues/179#issuecomment-898889925)? Could you re-open that issue? (Unfortunately I can't.)

Would you like to submit a PR to fix remaining issue(s)?

(sorry I fat-fingered and edited your comment first)

adams85:
@lahma I can take the other issue (the minor strict mode bug) but I'm not familiar enough with the parser to tackle this one.

lahma:
Seems that there's already [a work item](https://github.com/jquery/esprima/issues/1990) on Esprima side.

lahma:
By a cursory look it's enough two change:

`var isGenerator = isAsync ? false : Match("*");`

to 

`var isGenerator = Match("*");`

[...]

PR #189, opened 2021-08-19
Improvement on error reporting of 'for await' loops

adams85:
A minor fix related to async iterations (see https://github.com/sebastienros/esprima-dotnet/issues/183#issuecomment-901976754)


adams85:
Ooops, looks like this needs some more work:

https://github.com/sebastienros/esprima-dotnet/pull/189/commits/af79159487a33cd23a97eb7e49134fde324eb49d#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1R3045-R3055

Please don't merge this yet.

src/Esprima/JavascriptParser.cs
```diff
@@ -41,7 +41,7 @@ public Context()
             public bool AllowIn;
             public bool AllowStrictDirective;
             public bool AllowYield;
-            public bool Await;
```
lahma, line 44: even thought might not seem a good name, I think we should keep the naming that JS project has to make diffing and syncing easier

adams85: No problem, I'll revert it.

adams85:
@lahma Ok, I went over it again, I think it's ready now.

lahma:
I think we should wait for the Esprima change first, it also renames the `context.await` to `context.isAsync` which is more logical and will also reflect here (unfortunately clashing). I'd like to also see how this changes reported errors against JS test suite after those changes (ideally this would show improvements 😉 )

sebastienros:
Fixed the conflict from the GH UI, hopefully correctly.

lahma (approved):
I think it's all good, might need some styling adjustment later on but test cases are covering now 👍🏻

PR #191, opened 2021-08-21
AST polymorphism improvements (second thought)

adams85:
This is related to https://github.com/sebastienros/esprima-dotnet/pull/188.

Sorry for splitting hairs but in the meantime I realized that introducing a new interface for import specifiers is not really necessary. So I suggest getting rid of it because 
* having a base class and an interface for import specifiers seems a bit redundant,
* [as mentioned in the first PR](https://github.com/sebastienros/esprima-dotnet/pull/188#issuecomment-901861260), fields of AST nodes might be changed to properties in the future, in which case the `IImportDeclarationSpecifier` interface would become redundant anyway.

### DIFF
--- a/src/Esprima/Ast/IImportDeclarationSpecifier.cs
+++ b/src/Esprima/Ast/IImportDeclarationSpecifier.cs
@@ -1,8 +1,0 @@
-﻿namespace Esprima.Ast
-{
-    public interface IImportDeclarationSpecifier
-    {
-        Identifier Local { get; }
-        NodeCollection ChildNodes { get; }
-    }
-}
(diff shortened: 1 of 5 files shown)


lahma:
Are we nearing the moment that should prepare for 3.x release with proper getters to avoid usage of `new`, I'm not super exited with the hiding...

adams85:
@lahma 

>  I'm not super exited with the hiding

Yeah, it's ugly but unfortunately this is the only way to do it without introducing an interface and to avoid BC at the same time.

So it all comes down to this question: do you plan to release another 2.x version including the recent changes?

If so, then let's go with the hiding for now OR revert this change and postpone it until the fields vs. properties thing gets sorted out in 3.x. Just let's not introduce `IImportDeclarationSpecifier` in any circumstances because that would cause another probable BC in the future.

If not, then we can afford a BC, so let's just simply move the `Local` field to the base class.

I'll update the PR according to your answer.

lahma:
Let's go with this one for now, probably easier to break more after we get everything compatible with 2.x mindset merged 🙂

adams85:
Closes #180

PR #192, opened 2021-08-24
Allow assigning __proto__ in object literals

KurtGokhan:
Basically, this should work: `({ __proto__: [] })`

This should fail: `({ __proto__: [], __proto__: [] })`


lahma (approved):
Good catch, .NET version has the separate `BooleanValue` usage and indeed it was wrong in this case. Thanks for the investigation and the fix!

PR #193, opened 2021-08-29
Support ES2021 Logical Assignment

lahma:


lahma:
Checked on Jint side, parses and validates correctly,

PR #199, opened 2021-11-01
Work on a few extensions from esprim-next

jogibear9988:


jogibear9988:
I still need to work on the tests, I only merged the Code changes atm

jogibear9988:
@lahma @sebastienros 
should we rename "Import" class to "ImportExpression" as defined in EsTree Spec: https://github.com/estree/estree/blob/master/es2020.md#importexpression

jogibear9988:
@lahma @sebastienros 
is there a way to autimaticly update the expected test jsons? and also the test-262 file?

sebastienros:
> should we rename "Import" class to "ImportExpression" as defined in EsTree Spec:

Is it the only discrepancy?

lahma:
I think we should have a PR per feature, make it easier to validate the change and link commits/PRs between projects (what's our sync status). We should also make sure we have the same/equivalent PR in JS Esprima that has green build (merged or not) - esprima-dotnet syncs with esprima after all. 

I had a look at the private members thing and it seems that it has some copy paste and I think I had somewhat less repetitive solution, probably need to dig it up and check the status of that - may also be that mine was a worse hack though.


> is there a way to autimaticly update the expected test jsons?

Currently no.

> and do you not run the test-262 agains esprima-net?

Currently we trust the JS version suite and line-by-line porting, then validated in Jint that the feature works against actual interpreter.

jogibear9988:
I can create extra commits for each feature, but only if there is a chance they get merged.
I don't see that they will be in Esprima in the near future, as some of them are open there since over a year, are green, and don't get merged.
Also the last release of esprima is even longer ago (3years).
For the project I do, I'll need es6 classes support, so I'll wait if you will merge this, or live with a fork also of esprima.net until original esprima supports it.

jogibear9988:
ahh, I'll overread you'll wanted a green pull in esprima for each feature (merged or not).
so I'll split this pull up, and create a pull to esprima for the not existing ones

sebastienros:
These will definitely get merged. Be ready to get some comments about performance tweaks and test coverage though. Thanks a lot for all these contributions. If you feel like we are asking too much don't hesitate and let us know ;)

jogibear9988:
splitted in extra pull requests

lahma:
@jogibear9988 excellent! thanks for taking the extra time with this, we'll do our best to process these as fast as possible.

PR #201, opened 2021-11-02
Fix parsing error in exponent expressions with unary left-hand sides.

jogibear9988:
Esprima (correctly) rejects expressions like -1**2. - https://github.com/jquery/esprima/pull/2070
However, expressions like (-1)**2 are valid but still rejected.
This commit fixes this issue by identifying when the left operand
is parenthesized.

Fixes https://github.com/jquery/esprima/issues/1981


sebastienros:
Can you add the test case?

lahma (approved):
Looking good 👍🏻

PR #202, opened 2021-11-02
Fix ES2018 import()

jogibear9988:
https://github.com/jquery/esprima/pull/2076

### DIFF
--- a/test/Esprima.Tests/Fixtures/es2018/dynamic-import/await-import.tree.json
+++ b/test/Esprima.Tests/Fixtures/es2018/dynamic-import/await-import.tree.json
@@ -1,422 +1,148 @@
 {
-    "type": "Program",
-    "body": [
+  "type": "Program",
+  "body": [
+    {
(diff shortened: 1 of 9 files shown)


test/Esprima.Tests/Fixtures/es2018/dynamic-import/await-import.tree.json
```diff
@@ -1,422 +1,148 @@
...
-        }
-    ],
-    "sourceType": "script",
-    "range": [
-        0,
-        39
-    ],
-    "loc": {
-        "start": {
+  "type": "Program",
```
sebastienros, line 2: Is there a place I can find this file, or is it generated locally based on the test itself?

jogibear9988: Don't know what you mean...
The AST was different with the old implementation.... Now it is correct (I think :-) )

lahma (approved):
👍🏻

PR #204, opened 2021-11-02
BigInt & numeric separator support

jogibear9988:
there is not yet a pull request for esprima


src/Esprima/Scanner.cs
```diff
@@ -1084,6 +1051,27 @@ public bool IsImplicitOctalLiteral()
             return true;
         }
 
+        private string ScanLiteralPart(Func<char, bool> check)
+        {
+            string num = "";
+
+            if (Source.CharCodeAt(Index) == '_')
```
sebastienros, line 1058: I think the `Source.CharCodeAt(Index)` value should be saved locally to prevent multiple calls.

src/Esprima/Scanner.cs
```diff
@@ -1084,6 +1051,27 @@ public bool IsImplicitOctalLiteral()
             return true;
         }
 
+        private string ScanLiteralPart(Func<char, bool> check)
+        {
+            string num = "";
```
sebastienros, line 1056: We should prevent all these strings allocations by passing the outer StringBuilder.

jogibear9988: fixed

src/Esprima/Ast/BigIntLiteral.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Numerics;
+
+namespace Esprima.Ast
+{
+    public sealed class BigIntLiteral : Literal
+    {
+        public readonly string BigInt;
+
+        public BigInteger? BigIntValue => (BigInteger?) Value;
```
lahma, line 9: do we really need to add this as Literal already has that, I think it should be quite cheap to call even with the type check

src/Esprima/Scanner.cs
jogibear9988: For more perfomance, we could parse direcly into an int/long/double etc, without using a string. But I think that can be done in a another task

src/Esprima/Ast/BigIntLiteral.cs
jogibear9988: I don't need it, I only did to match the other...

jogibear9988: I did it to match the Tree spec: https://github.com/estree/estree/blob/master/es2020.md#bigintliteral

jogibear9988: Caus I think the AST should be nearly the same in all Javascript Parsers

src/Esprima/Scanner.cs
```diff
@@ -1084,6 +1051,27 @@ public bool IsImplicitOctalLiteral()
...
         }
 
+        private string ScanLiteralPart(Func<char, bool> check)
+        {
+            string num = "";
+
+            if (Source.CharCodeAt(Index) == '_')
+                ThrowUnexpectedToken(Messages.NumericSeperatorNotAllowedHere);
+
+            while (!Eof() && (check(Source.CharCodeAt(Index)) || Source.CharCodeAt(Index) == '_')) {
```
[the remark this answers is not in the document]
jogibear9988: maybe an issue with vs2022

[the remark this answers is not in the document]
lahma: might be, just a thing to keep in mind so code is kept consistent

jogibear9988:
is there anything I need to do here?

lahma:
I guess there's no Esprima PR to diff this against?

jogibear9988:
yes, I'll look when I've time to port my changes from esprima-next to esprima (should not be much work). but the development on esprima has nearly come to an halt, yes there were a few merges, but no reaction to questions, many dead branches... and no release since years...

jogibear9988:
pull req in esprima: https://github.com/jquery/esprima/pull/2105

lahma (approved):
Good to go, thanks for making the effort to get changes back to original JS project.

PR #205, opened 2021-11-02
support es2022 class properties & private properties

jogibear9988:
there is not yet a pull req for esprima


jogibear9988:
@lahma @sebastienros 
could you approve me, so the tests are run

jogibear9988:
https://github.com/jquery/esprima/pull/2106

PR #206, opened 2021-11-02
Fix build warnings

sebastienros:


sebastienros:
Also contains a `.gitattributes` to normalize line endings, as I believe this is causing issues with tests failings on my machine.
Applied `dotnet format`.

Issue #157, opened 2021-05-10
Parse from JSON?


ejsmith:
I'm looking to build a GUI builder for conditions and simple statements. It seems like I would want to make the GUI build up an ES tree and persist that tree for when I need to go back and re-populate the visual builder UI. I don't want to create my own tree model from scratch so I'd like to be able to serialize the model to JSON which I can do using `ToJsonString`, but it doesn't appear that there is a way to parse that JSON back into the ES tree models. Is that correct? Any ideas on how I would go about doing this?

jogibear9988:
@ejsmith here is a sample converter wich could be used with System.Text.Json, but it's not yet finished.
We need to add all the ast nodes, make ist error tolerant and so on...
https://github.com/sebastienros/esprima-dotnet/pull/207

PR #207, opened 2021-11-04
example converter from json to ast

jogibear9988:
fix for #157


jogibear9988:
@sebastienros @lahma 
what do you think about this?
Is this something you want in the lib, so I could look to finish it.
Or do we not need it?

lahma (commented):
I think the general idea is sound. Supporting new .NET feels like a good choice, of course until someone complaints about not having this on their Windows XP or something.

PR #208, opened 2021-11-06
Create test bed to test against ECMAScript 262 files

lahma:
Idea here is to mimic JS version's testing and create a file showing errors/missing bits so we can detect regressions etc.

* uses `Test262Harness` package to download and run tests, `Program.cs` has hard-coded GitHub hash to download and test against
* there's now `allow-list.txt` which is our TODO
* GH actions check that allow list and current failing/succeeding match
* file can be re-generated when needed when fixing/implementing features, diff should always be removing lines


.github/workflows/pr.yml
```diff
@@ -53,5 +60,17 @@ jobs:
       with:
         dotnet-version: 5.0.*
 
+    - name: Setup Node and 262 test suite
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14.x
```
sebastienros, line 66: Why not newer version with package lock files v2?

test/Esprima.Tests.Test262/Esprima.Tests.Test262.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net5.0</TargetFramework>
+        <Nullable>enable</Nullable>
+        <OutputType>Exe</OutputType>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Spectre.Console" Version="0.42.0" />
```
sebastienros, line 10: See that you decided to have some fun too ;)

.github/workflows/pr.yml
lahma: I'm just copy-pasting here from the past, I can check to update this one

test/Esprima.Tests.Test262/Esprima.Tests.Test262.csproj
lahma: There has to be some incentive to do such laborous tasks! A fun experiment to get red/green status on screen without worrying about creating new dependencies for consumers at least...

lahma: It's kinda cool though:

image

sebastienros: Is that just locally or does it work also in the GH console?

lahma: It should revert to least cool experience on non-interactive console, experimenting here...

lahma:
I think I'm going to extract the harness and parsing to separate project so I can also test against different JS engines.

jogibear9988:
In esprima-next im now down to 4818 not workin tests from 262.
Where in original esprima fail over 11000

.github/workflows/pr.yml
lahma: Now using logic to download repository archive, so long package.json and whatnot.

.editorconfig
```diff
@@ -12,7 +12,6 @@ indent_style = space
 [*.{cs,csx,vb,vbx}]
 indent_size = 4
 insert_final_newline = true
-charset = utf-8-bom
```
lahma, line 15: The opinion seems to be that BOM is redundant and doesn't help much, so removed.

sebastienros:
Is there an argument to list the files with issues?

lahma:
allow-list.txt is the list 🙂

lahma:
Also, the two last commits show how build breaks when failing file is not in allow-list.txt.

sebastienros:
Ah ok, like skipped actually.

sebastienros (approved):

PR #211, opened 2021-11-21
Ast visitor Refactoring

jogibear9988:
WIP


jogibear9988:
@sebastienros @lahma 
good idee, about how to created this wethod?
https://github.com/sebastienros/esprima-dotnet/pull/211/commits/a54af62b3e242dcf43fccf072b28ec2eb198d129#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR17

I'd like to use a ref retrun and ref in but if I use a RefReturn, I could not return a new NodeList.
so whats the best way to do it?

jogibear9988:
There are also a few more Questions:

We have following classes, wich have no own Visit Method, but this Classes have extra Properties:

BigInt
ClassProperty
Directive
Module
Script

We have the following ones wich have no extra Information, but also no own Visit Method:

BindingPattern
ComputedMemberExpression
Declaration
ExportDeclaration
Expression -> stamentListItem
ImportDeclarationSpecifier
Statement
StaticmemberExpression

And we have a Special case in
BinaryExpression, where we visit the Binary Expression as Logical with no extra Derived Type.

so, what sould we do abaout all those cases? Should we change something? Leave it as is?

adams85:
@jogibear9988 I think the ideal situation would be if only concrete (non-abstract) node types had a corresponding `VisitXXX` method. Unfortunately the original visitor implementation wasn't designed this way and I didn't want to introduce BCs when I was updating it. But now we're going to introduce a BC anyway (by changing the return type of `AstVisitor.Visit`), so maybe this is the right time to set this straight.

* BigInt: this one right away goes against the rule of thumb above  as it has a non-abstract base class (`Literal`), which already has its corresponding `VisitLiteral` method. `BigIntLiteral` has no child nodes to visit though, so we may leave this as is, i.e. I wouldn't introduce `VisitBigIntLiteral`. Extra property can be accessed in `VisitLiteral` after checking type and downcasting.

* ClassProperty: this one is abstract and all node types deriving from it (`MethodDefinition`, `Property`, `PropertyDefinition`) have their corresponding `VisitXXX` methods. I wouldn't touch this.

* Directive: situation is similar to BigInt.

* Module, Script: according to our rule of thumb this should be changed: `VisitProgram` should be replaced by `VisitModule` and `VisitScript`. But even in this case we wouldn't gain much, so I'm not sure it's worth the hassle.

The case of BindingPattern, Declaration, ExportDeclaration, Expression, ImportDeclarationSpecifier, Statement are like ClassProperty, while that of ComputedMemberExpression, StaticMemberExpression is similar to BigInt.

Bottom line, I don't see much room for improvement here. However, go for it if you do. I think changes are acceptable as long as the following two criteria are met:
* Every node is visited once and only once.
* Every visitation goes through `AstVisitor.Visit` (i.e. no shortcuts).

One more remark: I think it'd be better to extract rewriting/updating of the visited nodes into separate methods as it's done in the framework's expression tree visitor (see [1](https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/ExpressionVisitor.cs#L232) and [2](https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/ConditionalExpression.cs#L90)). This would make the life of inheritors of `AstVisitor` easier when they need to override some `VisitXXX` method (e.g. less code to copy). I'm not sure though that these `Update`/`Rewrite` methods should go into the node classes. If they were placed in `AstVisitor` as virtual methods, inheritors who don't want to change the tree, just traverse it, could override it to simply return the visited node.

jogibear9988:
@adams85 sorry, I did not look that some of the Types were abstract, my fault.

But do you also have a good Idea about my first question. On how to create the NodeVisitor without copy of the NodeList?

adams85:
I assume that you have issues with this line:
https://github.com/sebastienros/esprima-dotnet/pull/211/commits/a54af62b3e242dcf43fccf072b28ec2eb198d129#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR44

You can't really return a local variable by reference as its lifetime is tied to the declaring method. After the method has returned, it doesn't exist any more.

You can do something like this:
``` csharp
protected internal virtual bool VisitNodeList<T>(in NodeList<T> nodes, out NodeList<T> newNodes) where T : Node
{
    // ...omitted for brevity...

    if (newNodeList != null)
    {
        newNodes = new NodeList<T>(newNodeList);
        return true;
    }

    newNodes = default;
    return false;
}
```

And then:

``` csharp
protected internal virtual Node? VisitFunctionDeclaration(FunctionDeclaration functionDeclaration)
{
    Identifier? id = null;
    if (functionDeclaration.Id is not null)
    {
        id = (Identifier?) Visit(functionDeclaration.Id);
    }

    var paramsChanged = VisitNodeList(in functionDeclaration.Params, out var newParams);
    var body = Visit(functionDeclaration.Body);

    if (functionDeclaration.Id != id || paramsChanged || functionDeclaration.Body != body)
    {
        if (body == null)
        {
            return null;
        }
        return new FunctionDeclaration(id, in newParams, (BlockStatement) body, functionDeclaration.Generator, functionDeclaration.Strict, functionDeclaration.Async);
    }
    return functionDeclaration;
}
```

adams85:
BTW, this looks suspicious:
https://github.com/sebastienros/esprima-dotnet/commit/a54af62b3e242dcf43fccf072b28ec2eb198d129#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR30-R37

Shouldn't the new list contain the original parameters from 0 to i-1?

```csharp
    newNodeList = new List<T>(nodes.Count);
    for (var j = 0; j < i; j++)
    {
        newNodeList.Add(nodes[i]);
    }
```

And a bit of nitpicking: I suggest using pattern matching for null checks (`is null` instead of `!= null`).

jogibear9988:
> ```cs
> for (var j = 0; j < i; j++)
>     {
>         newNodeList.Add(nodes[j]);
>     }
> ```

yes, I forgott it... :-)

jogibear9988:
@adams85 maybe take a look at my new commit...
If done the Update as an extra method for VisitModule, and in one for VisitProgram.
Do you think this a good way to go?

adams85:
@jogibear9988 I'm in favor of `VisitModule` + `UpdateModule`. As a matter of fact, I think we may consider providing a "read-only" (traverse-only) visitor as well. What if we did something like this:

```csharp
public class AstVisitor
{
    protected virtual bool VisitNodeListAndIsNew<T>(in NodeList<T> nodes, out NodeList<T> newNodes) where T : Node
    {
        newNodes = default;
        return false;
    }

    protected internal virtual Node? VisitModule(Module module)
    {
        var newParamsList = VisitNodeListAndIsNew(in module.Body, out var newStatements);
        return UpdateModule(module, newParamsList, ref newStatements);
    }

    protected virtual Module? UpdateModule(Module module, bool newParamsList, ref NodeList<Statement> newStatements)
    {
        return module;
    }

    // rest is omitted for brevity...
}

public class AstMutator : AstVisitor
{
    protected override bool VisitNodeListAndIsNew<T>(in NodeList<T> nodes, out NodeList<T> newNodes) where T : Node
    {
        // implementation which can handle changes...
    }

    protected override Module? UpdateModule(Module module, bool newParamsList, ref NodeList<Statement> newStatements)
    {
        // implementation which can handle changes...
    }
}
```

By doing this, traverse-only visitors could keep nearly the same performance as before since they don't have to pay the price of extra checks needed for modification. (Only some extra virtual calls are done.) What's your opinion on this?

BTW, `VisitNodeListAndIsNew` is a bit weird. I know naming things is hard but maybe we can come up with something better. How about `HasNodeListChanged` or `CheckNodeListChanged`? Or another option is to revert the name to `VisitNodeList` and switch the return value and the out parameter like this: `NodeList<T> VisitNodeList<T>(in NodeList<T> nodes, out bool hasChanged)`.

jogibear9988:
sorry thias takes so a long time, but I'm busy at work. Will finish this next week when my holiday starts

adams85:
Ok, no rush.

lahma:
I think most of the stuff has been done? Or there's a great merge conflict ahead...

jogibear9988:
for me it's okay

adams85:
As I see, everything this PR wanted to achieve has been done already. See
* https://github.com/sebastienros/esprima-dotnet/pull/243
* https://github.com/sebastienros/esprima-dotnet/pull/253
* https://github.com/sebastienros/esprima-dotnet/pull/257

As for the relationship between AST nodes and `VisitXXX` methods, there are some node types which are still handled polymorphically:
* `Module`, `Script` -> `VisitProgram`
* `Directive` -> `VisitExpressionStatement`
* `ComputedMemberExpression`, `StaticMemberExpression` -> `VisitMemberExpression`
* `UpdateExpression` -> `VisitUnaryExpression`
* `StaticBlock` -> `VisitBlockStatement`
* `BigIntLiteral` -> `VisitLiteral`

However,I think the situation is good enough, I wouldn't make further changes.

So, if you agree, this PR can be canceled.

lahma:
Closing per suggestion, maybe a new PR for something specific still missing. Thanks for laying the foundation here!

PR #214, opened 2021-12-13
Fix BigInt parsing logic

lahma:
* fix various issues found while integrating into Jint
* add netstandard2.1 target
* use spans for literal parsing
* add System.Memory as dependency


src/Esprima/Scanner.cs
```diff
@@ -947,12 +955,34 @@ public Token ScanHexLiteral(int start)
...
 
+        private Token ScanBigIntLiteral(int start, ReadOnlySpan<char> number, NumberStyles style)
+        {
+            var c = number[0];
+            if (c >= 8 && c <= 'F')
+            {
+                // ensure we get positive number
+                number = ("0" + number.ToString()).AsSpan();
+            }
+#if !NETFRAMEWORK && !NETSTANDARD2_0
```
sebastienros, line 966:
```suggestion
#if NETSTANDARD2_1_OR_GREATER
```

Is that what you meant?

src/Esprima/Esprima.csproj
```diff
@@ -26,12 +26,14 @@
...
-    
+
     <DefineConstants Condition="'$(Checked)' == true or $(Configuration) == 'Debug'">$(DefineConstants);LOCATION_ASSERTS</DefineConstants>
-    
+
   </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="System.Memory" Version="4.5.4" />
```
sebastienros, line 36: 1- What's coming from it?
2- Would it be removed if TFMs included netcoreapp3.1 ... net6.0?

src/Esprima/Scanner.cs
lahma: Does it include everything like NETCOREAPP31 onwards? standard vs modern net5/6 can be confusing.

src/Esprima/Esprima.csproj
lahma: Basically the AsSpan, might  be good idea to refine targets and remove dep if possible

sebastienros:
Can you add the `new` operator on `BigIntLiteral.cs` while you are at it, to remove the warnings?

src/Esprima/Esprima.csproj
```diff
@@ -26,12 +26,15 @@
...
+
     <DefineConstants Condition="'$(Checked)' == true or $(Configuration) == 'Debug'">$(DefineConstants);LOCATION_ASSERTS</DefineConstants>
-    
+
+    <DefineConstants Condition=" '$(TargetFramework)' == 'netstandard2.1'">$(DefineConstants);HAS_SPAN_PARSE</DefineConstants>
   </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="System.Memory" Version="4.5.4" Condition=" '$(TargetFramework)' != 'netstandard2.1'"/>
```
sebastienros, line 37: Is it required for netstandard2.1 ?

src/Esprima/Scanner.cs
```diff
@@ -947,12 +955,34 @@ public Token ScanHexLiteral(int start)
             };
         }
 
+        private Token ScanBigIntLiteral(int start, ReadOnlySpan<char> number, NumberStyles style)
+        {
+            var c = number[0];
+            if (c > 7 && Character.IsHexDigit(c))
+            {
+                // ensure we get positive number
+                number = ("0" + number.ToString()).AsSpan();
```
sebastienros, line 964: ToString() on spans, maybe this method should just take the original string, that had to be converted as a span.

src/Esprima/Esprima.csproj
lahma: Seems that compilation works without it.

src/Esprima/Scanner.cs
lahma: Span is part of read source, this is a special case when we need to materialize in order to get parsing to work correctly.

lahma:
> Can you add the `new` operator on `BigIntLiteral.cs` while you are at it, to remove the warnings?

Pushed a commit.

sebastienros:
Not this PR, I feel like we need a custom class for each Literal type. Would also prevent boxing.

lahma:
I'm fairly confident that now everything in this PR is OK, I have two failing test cases on Jint side but I hope they are unrelated to parsing anymore.

sebastienros (approved):

PR #216, opened 2021-12-17
Fix parsing binary number with BigInt postfix

lahma:
Found another parsing case missed with broader test suite: `0b00n`, a small binary number with `n` that means that it should be parsed as BigInt.


src/Esprima/Scanner.cs
```diff
@@ -990,11 +990,16 @@ public Token ScanBinaryLiteral(int start)
                 ThrowUnexpectedToken();
             }
 
+            var bigInt = number.Length > 4;
```
sebastienros, line 993: I did some test in chrome and it doesn't seem to assume binary literals with more than 4 digits are bigints. Is that in the spec?

sebastienros: I can only get a bigint if I add the n suffic

lahma: indeed this was incorrect, the parse was called from BigInt constructor which itself needs to do these conversion for *strings* containing this textual value. I already had handling for hex and octal in there - tired brain I guess. Removed the length check.

PR #217, opened 2021-12-19
Fix module import/export identifier parsing

lahma:
[Per specification](https://tc39.es/ecma262/#prod-ModuleExportName), module `export`s/`import`s can either have `Identifier` or `StringLiteral` as value. 

This PR fixes the issue. Is a breaking change as API changes from `Identifier`s to `Expression`s, but there's no other way to fix the issue. This PR unblocks failing tests in Jint PR.

### DIFF
--- a/src/Esprima/Ast/ExportAllDeclaration.cs
+++ b/src/Esprima/Ast/ExportAllDeclaration.cs
@@ -5,13 +5,17 @@ namespace Esprima.Ast
     public sealed class ExportAllDeclaration : ExportDeclaration
     {
         public readonly Literal Source;
-        public readonly Identifier? Exported;
+
+        /// <summary>
+        /// Identifier | StringLiteral
+        /// </summary>
+        public readonly Expression? Exported;
 
         public ExportAllDeclaration(Literal source) : this(source, null)
         {
         }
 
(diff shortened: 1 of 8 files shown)


jogibear9988:
@lahma is this something I also need to change in esprima-next?

lahma:
@jogibear9988 yes I think esprima-next node definitions should also allow `StringLiteral` in these node types, see `Identifier | StringLiteral` comments in this PR. I think you can validate this with test suite parsing code, whitelist.txt problems should shrink with these changes.

jogibear9988:
@lahma 
If I look in the estree spec, i found only "Identifier", also in typescript my imports do not work if I change them to strings
https://github.com/estree/estree/blob/master/es2015.md#importspecifier
Where is the spec?

lahma:
[This was the one](https://tc39.es/ecma262/#prod-ModuleExportName) I referred to. So these are valid too:

```
export { "☿" as Ami } from "./export-expname_FIXTURE.js"
import { "☿" as Ami } from "./export-expname_FIXTURE.js"
```

[See AST explorer](https://astexplorer.net/#/gist/0e7f37e4e8d5d04aacae4f640b66e57c/05134d421289162e3b9fea6f56f5e75abac82871).

jogibear9988:
@lahma 
it's now also fixed in estree spec:
https://github.com/estree/estree/blob/master/es2022.md#importspecifier

PR #222, opened 2022-01-03
Support class static blocks

jogibear9988:
commit b6ec3c79d317ed0d8d861c72567730a0f723ba3e from esprima-next - parse class "static blocks"

### DIFF
--- a/src/Esprima/Ast/BlockStatement.cs
+++ b/src/Esprima/Ast/BlockStatement.cs
@@ -2,11 +2,16 @@
 
 namespace Esprima.Ast
 {
-    public sealed class BlockStatement : Statement
+    public class BlockStatement : Statement
     {
         private readonly NodeList<Statement> _body;
 
         public BlockStatement(in NodeList<Statement> body) : base(Nodes.BlockStatement)
+        {
+            _body = body;
+        }
+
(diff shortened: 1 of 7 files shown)


jogibear9988:
@lahma 
breaking change, it changes list type of class body.
in estree spec it is a or of 3 different types, this is not possible in C#
https://github.com/estree/estree/blob/master/es2022.md#classbody

lahma (approved):

lahma:
Seems that it has to break compatibility, thanks for this.

jogibear9988:
@lahma 
Should we also change https://github.com/sebastienros/esprima-dotnet/blob/cf68c4db7a3548de83d64d4dc678c446346fd248/src/Esprima/Ast/MethodDefinition.cs#L5
According to definition from EsTree it is derived from Node and not Class Property.
We then need to copy all Properties from ClassProperty

lahma:
I also noticed that currently `PrivateIdentifier` is derived from `Expression` and in spec it's said to be deriving from `Node` maybe these both could be changed? Unless there are even more I haven't noticed yet. With these I think we can either target 2.2 or even 3.0, there's also other cleanup planned later I guess.

PR #228, opened 2022-02-20
Fix update-allow-list logic, update test262harness

lahma:
* Now can run project and update allow list from command line, tests should still work per usual
* Updated new version of test262harness that fixes negative test case handling (this parse should have thrown an error)
* Updated allow-list to contain new problems where test case should throw, but didn't
* Also running allow list check as part of build, it will return non-zero exit code if allow-list is out of date


.github/workflows/pr.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     - name: Generate test cases
       working-directory: ./test/Esprima.Tests.Test262
-      run: dotnet tool restore && dotnet test262 generate
+      run: dotnet tool restore && dotnet test262 generate && dotnet run -c Release -- --update-allow-list
```
sebastienros, line 49: If that updates the allow-list, then this is just saying "anything that doesn't work is ignored", right? So if I create a PR that only breaks a unit test, would it just be ignored and hide it from us?

sebastienros: > Also running allow list check as part of build, it will return non-zero exit code if allow-list is out of date

PR #232, opened 2022-03-27
Fix some numeric parsing related issues

lahma:
First tried to fix some numeric separator issues but started to be bit of yak shaving. 

Found out that the tolerant logic in options was wrong, tests should be run with Tolerant = false to enforce proper parsing rules. Changed that and did required fixes. Options are still Tolerant = true by default as it would be a breaking change to change that.

Some errors are easiest to handle on Scanner level so supplied more information to Scanner for it to know with error should be thrown (are we in strict mode or not, are we allowing escaped identifier names like for object properties which is allowed).

Deleted couple old test cases which are not meant to throw based on my nodejs testing.

allow-list.txt should be a good indicator for changes, almost 2000 lines removed, some lines added like top level await (which hasn't been implemented).


sebastienros (approved):
the allow-list is surprisingly different

PR #234, opened 2022-04-12
Replace hasProto token with ref bool

lahma:
The old code inherited from esprima used a token to pass info about proto being present to avoid JS's copy by value semantics, in C# it makes more sense just just ref bool.


sebastienros:
Can't approve a merged PR, weird

lahma:
Hehe, I thought this was so simple that even I couldn't screw it up...

PR #239, opened 2022-04-24
AstVisitor corrections

adams85:
I've found two minor issues related to `AstVisitor` in the recent commits:

1. `VisitExportAllDeclaration` was not updated by #196 to include the visitation of the new `ExportAllDeclaration.Exported` property.
2. Addition of `VisitStaticBlock` is rather questionable because it adds redundancy (at least, from the visitor's POV). `StaticBlock` is just a special case of `BlockStatement`, has no additional attributes. In such cases we (usually) haven't introduced dedicated visit methods, just left it to the base class to handle the visitation aspect. For example, in the cases of `ExpressionStatement` and `Directive` or `MemberExpression` and `StaticMemberExpression`. I think it'd be a good idea to keep up this convention.

### DIFF
--- a/src/Esprima/Ast/StaticBlock.cs
+++ b/src/Esprima/Ast/StaticBlock.cs
@@ -1,16 +1,9 @@
-﻿using Esprima.Utils;
-
-namespace Esprima.Ast
+﻿namespace Esprima.Ast
 {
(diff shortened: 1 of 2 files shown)


lahma (approved):
Looks good.

PR #243, opened 2022-05-05
Astvisitor updated for changing ast

Xicy:
New implementation started according to PR #211


test/Esprima.Tests/ConverterTests.cs
```diff
@@ -12,14 +12,21 @@ public class ConverterTests
...
     }
 
     private static T? FindNearTypeOfDescend<T>(Node node) where T : Node
     {
         return node.DescendantNodesAndSelf().OfType<T>().FirstOrDefault(_ => true, null);
     }
 
+    private static object? FindNearTypeOfDescendTyped(Type type, Node node)
+    {
+        return typeof(ConverterTests)
```
sebastienros, line 25: Don't use reflection here. Check the type in linq directly. The generic method can then call this one.

Xicy:
Example;

Input
```js
var React = require('/react'); return (<div>Test</div>);
```

Output
```js
var React = require('/react');
return React.createElement('div', null, ['Test']);
```

Converter Usage

```csharp
    var parserOptions = new ParserOptions()
    {
        AdaptRegexp = true,
        Tolerant = true,
        Jsx = true
    };
        
    var script = new JavaScriptParser(@"var React = require('/react'); return (<div>Test</div>);", parserOptions).ParseScript();
    
    JSXConverter _jsxConverter = new();
    var convertedScript = _jsxConverter.Apply(script);
    var asJson = convertedScript.ToJsonString();
```

JSX Converter Implementation

```csharp

internal sealed class JSXConverter : AstConverter
{
    private static readonly MemberExpression ReactCreateElement = new StaticMemberExpression(new Identifier("React"), new Identifier("createElement"), false);
    private static readonly Func<string,MemberExpression> ObjectCaller = (property) => new StaticMemberExpression(new Identifier("Object"), new Identifier(property), false);
    private static Regex key = new("/^[a-zA-Z$][\\w$]*$/");
    private static Regex match = new("/^[ \\t]*[\\r\\n][ \\t\\r\\n]*$/");
    private static Regex left = new("/^(?:[\\r\\n]+\\s*)+/");
    private static Regex right = new("/(?:\\s*[\\r\\n]+)+$/");
    
    private Node? Root;

    public T? Apply<T>(T node) where T:Node
    {
        Root = null;
        return Visit(node) as T;
    }
    
    public override Node? Visit(Node? node)
    {
        Root ??= node;
        
        switch (node)
        {
            case JsxElement element:
            {
                var openingElement = (element.OpeningElement as JsxOpeningElement)!;
                var args = NodeList.Create(new[]{ ParseCallee(openingElement.Name), ParseArg(openingElement.Attributes), ParseChildren(element.Children) });
                var reactCreateElementInstance = new CallExpression(ReactCreateElement, args, false);
                return base.Visit(reactCreateElementInstance);
            }
            case JsxExpressionContainer expressionContainer:
                return Visit(expressionContainer.Expression);
            case JsxText text:
                return new Literal(text.Value, text.Raw);
            default:
                return base.Visit(node);
        }
    }

    private Expression ParseArg(NodeList<JsxExpression> attributes)
    {
        var argsForObjectAssign = new List<Expression>();
        List<Expression> attrs = new();
        foreach (var attribute in attributes)
        {
            switch (attribute)
            {
                case JsxSpreadAttribute spreadAttribute:
                {
                    if (attrs.Any())
                    {
                        argsForObjectAssign.Add(new ObjectExpression(NodeList.Create(attrs)));
                        attrs = new();
                    }
                    argsForObjectAssign.Add(spreadAttribute.Argument);
                    continue;
                }
                case JsxAttribute jsxAttribute:
                {
                    var name = jsxAttribute.Name;
                    var value = jsxAttribute.Value ?? (name is JsxIdentifier i ? new Identifier(i.Name) : name);
                    attrs.Add(new Property(PropertyKind.Init,toKey(name),false,value,false,false));
                    break;
                }
                default:
                    throw new NotImplementedException();
            }

            Expression toKey(Expression exp)
            {
                var name = exp switch
                {
                    JsxIdentifier jsxIdentifier => jsxIdentifier.Name,
                    Identifier identifier => identifier.Name,
                    _ => throw new NotImplementedException()
                };

                return key.IsMatch(name) ? Reuse(exp) : new Literal(name);
            }
            
        }

        if (attrs.Any())
        {
            argsForObjectAssign.Add(new ObjectExpression(NodeList.Create(attrs)));
        }

        switch (argsForObjectAssign.Count)
        {
            case 0:
                return new Literal("null");
            case 1:
                return argsForObjectAssign[0];
            case > 1 when argsForObjectAssign[0] == attributes[0]:
            {
                var args = NodeList.Create(new []{ argsForObjectAssign[0] });
                argsForObjectAssign[0] = new CallExpression(ObjectCaller("create"), args, false);
                break;
            }
        }

        return new CallExpression(ObjectCaller("assign"), NodeList.Create(argsForObjectAssign), false);
    }
    
    private Expression ParseChildren(NodeList<JsxExpression> children)
    {
        var allExpressions = children.Where(x => x is not JsxText || (x is JsxText text && !match.IsMatch(text.Value))).ToList();
        var end = allExpressions.Count - 1;
        { allExpressions[0] = allExpressions[0] is JsxText text ? new JsxText(left.Replace(text.Value, ""), text.Value) : allExpressions[0]; }
        { allExpressions[end] = allExpressions[end] is JsxText text ? new JsxText(right.Replace(text.Value, ""), text.Value) : allExpressions[end]; }
        var expressionList = allExpressions.Select(Visit).OfType<Expression>().ToList();
        return new ArrayExpression(NodeList.Create(expressionList));
    } 
    
    private Expression ParseCallee(JsxExpression expression)
    {
        var target = GetTarget(expression);
        bool StatementFinder(Node y) => y is VariableDeclaration v && v.Declarations.Any(z => z.Id is Identifier i && ((target is JsxMemberExpression e && i.Name == toStr(e.Object)) || (target is JsxIdentifier ii2 && i.Name == ii2.Name) || (target is Identifier ii && i.Name == ii.Name)));
        var isContextHasDeclaration = target.AncestorNodes(Root).Any(x => x.ChildNodes.Any(StatementFinder)); 

        if (isContextHasDeclaration)
        {
            return Reuse(target);
        }

        var literal = toStr(target);
        return new Literal(literal,literal);

        string? toStr(Expression expr)
        {
            return expr switch
            {
                JsxMemberExpression e => toStr(e.Object) + "." + toStr(e.Property),
                JsxIdentifier id => id.Name,
                Identifier id => id.Name,
                _ => throw new NotImplementedException()
            };
        }
        
    }

    private Expression GetTarget(Expression expression)
    {
        return expression switch
        {
            JsxMemberExpression memberExpression => GetTarget(memberExpression),
            JsxNamespacedName namespacedName => namespacedName.Name,
            JsxIdentifier identifier => identifier,
            _ => throw new NotImplementedException()
        };
    }

    private Expression Reuse(Expression expression)
    {
        return expression switch
        {
            JsxMemberExpression memberExpression => new ComputedMemberExpression(Reuse(memberExpression.Object), Reuse(memberExpression.Property), false),
            JsxIdentifier identifier => new Identifier(identifier.Name),
            JsxNamespacedName namespacedName => new ComputedMemberExpression(namespacedName.Namespace, namespacedName.Name, false),
            _ => expression
        };
    }
}
```

CharlieEriksen:
This PR is awesome. I've been trying it out, and it's great. Can't wait to see it in a full release :)

lahma:
PInging @jogibear9988 and @adams85 too for a review, great candidates for doing a peer review as both have had their share of AST traversing 😉

jogibear9988:
LGTM

adams85:
Ok, this PR is quite a bit, so I'll need some time to go through it.

At the moment I have two remarks:
1. I suggest rebasing instead of merging with the main branch. Keeps the PR easier to follow. Squashing into some commits would be useful too.
2. Is JSX part of the JS standard now? Excuse my ignorance, I'm not completely up-to-date with the developments of the JS world... However, if it's not, I don't think it's a good idea to bake JSX-related things into the core library. Shouldn't it be some kind of extension to it? Like providing some extension points in the core library so that this can be implemented in another package called something like `Esprima.Jsx`? As far as I can see, there are use cases (including mine) where JSX is irrelevant and handling its aspects (like visitation) creates worthless additional work. @sebastienros, @lahma, @jogibear9988 What are your thoughts on this?

lahma:
And now while I still remember.. I've changed esprima builds to to deliver 3.0 betas (beta1 on NuGet) so now is the time to move and change things (like AST API). Jint is dogfooding and idea is to clean up things and fix/implement missing bits as much as possible before stable release at some point. MyGet serves every main build as usual and occasional releases to NuGet when needed.

Xicy:
Thank you for all comments,

I've will try to answer some of them, Excuse if I wrong understand, please keep in mind to correct to me.

@adams85 
1- I never used before Rebase, I researched now. As I understand I had to made this instead of merge so should I do something for that ?
2- As I understand I should not do any thing for a now , as @lahma 's comments. I can work in soon to refactor it with `JsxParser extends JavasScriptParser` logic.

@CharlieEriksen 
I had forgot the why I design with [ExpressionType], so I tried again now. I faced again some issue, method has to TypeConversion, I've try to explain with example in the below. You can look in comments in the code. If any one have idea about to Node types, I can try it.

```csharp
// Updated via Node? type
    protected override Node? UpdateProgram(Program program, bool isNewStatements, ref NodeList<Statement> statements)
    {
        if (!isNewStatements)
        {
            return program;
        }

        return program switch
        {
            Module => new Module(statements /* NodeList<Statement> */),
            Script script => new Script(statements /* NodeList<Statement> */ , script.Strict),
            _ => throw new NotImplementedException($"{program.SourceType} does not implemented yet.")
        };
    }
```

```csharp
    protected internal virtual Node? VisitProgram(Program program)
    {
        // Actual problem is in there 'HasNodeListChanged'. When I used to Node type 
        // instead of T type is create NodeList<Node> and I couldn't use to create new Program. 
        // Because ast node takes self NodeList<T> type, for that example Module and Script takes 
        // NodeList<Statement>
        var isNewStatements = HasNodeListChanged(program.Body, out var statements);
        return UpdateProgram(program, isNewStatements, ref statements);
    }
```

CleanShot 2022-05-30 20 31 15

lahma:
> I agree. This PR is beefy enough already, so we'd better deal with separation of JSX later, after this PR has been merged. But I think it should be sorted out before v3.0 is released. @lahma Is this what you mean by "so now is the time to move and change things (like AST API)"? Anyway, I'm going to open an issue to discuss this matter soon.

We can clean up old API and improve, hide things that have been accidentally too public etc. I for one have a PR in works to change token to struct (if it brings better perf that is). There was a mix of properties and public fields. NET 6 runtime is quite good at optimizing so we could consider going the get-property route for consistency, maybe some `ReadOnlySpan<T>` usage in public API, maybe an umbrella issue/discussion would indeed be good to have.

Xicy:
@adams85 
>It would be nice to clean up this PR but due to the merges done already, I'm afraid it wouldn't be an easy task now. However, in your future PRs, please, use rebasing if possible. Here's [an article](https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request) which can help you to grasp the idea.

Thank you for this information it is great experience to learn.

>@Xicy Am I missing something here? Sorry, I couldn't completely follow your example. Could you elaborate a bit further on this?

Maybe My test are broke, I'm not sure about that. I shared a new branch (https://github.com/Xicy/esprima-dotnet/tree/astvisitor-node) this applied to node changes. Also please to try it with below example.

adams85:
Okay, I checked out your new branch and tested your sample code.

It throws an `InvalidCastExpression` on [this line](https://github.com/Xicy/esprima-dotnet/blob/2500fbe9bb30d35b6bada2b09feaaf5c50835d98/src/Esprima/Utils/AstConverter.cs#L36). At this point you're visiting and updating a list of `JsxExpression` nodes in a call to `HasNodeListChanged<JsxExpression>`. [Here](https://github.com/Xicy/esprima-dotnet/blob/2500fbe9bb30d35b6bada2b09feaaf5c50835d98/src/Esprima/Utils/AstVisitor.Jsx.cs#L37) is where this call happens.

In this `HasNodeListChanged<JsxExpression>` call you visit the elements of a `NodeList<JsxExpression>` but there's a visitation which doesn't return a `JsxExpression` but a `Literal`. And you obviously can't add a `Literal` to a `NodeList<JsxExpression>`.

The cause of the problem lies in your sample code, which indeed converts a `JsxExpression` (`JsxText`, precisely) to a `Literal`:

```csharp
    protected override Node? UpdateJsxText(JsxText jsxText)
    {
        return new Literal(jsxText.Value, jsxText.Raw);
    }
```

To me this doesn't look like a shortcoming of the library but of the consuming code.

I can imagine that in some situations converting a `JsxText` to a `Literal` makes sense but in other situations it doesn't. In other words, in some cases we could restrict the return type of `UpdateJsxText` to e.g. `JsxExpression` but I'm not sure we could restrict it in *all* cases. As far as I can see, only the consumer knows what they want to do with the AST, so it's their responsibility to return the correct node type. I may be wrong though. What do you think?

CharlieEriksen:
@adams85 

Here's a few samples of what I'm doing:

```csharp
public class EmptyStringConditional: AstConverter
{
    public override Node? Visit(Node? node)
    {
        if (node is BinaryExpression
            {
                Operator: BinaryOperator.LogicalOr,
                Right: Literal {StringValue:""},
                
            } ce)
        {
            return ce.Left;
        }

        return base.Visit(node);
    }
}
```

```csharp
public class CombineBinaryLiterals : AstConverter
{
    public override Node? Visit(Node? node)
    {
        if (node is BinaryExpression b && b.Left is Literal left && b.Right is Literal right)
        {
            var value = (left.StringValue ?? left.Raw) + (right.StringValue ?? right.Raw);
            var newLiteral =  new Literal(value, $"\"{value}\"" );
            return base.Visit(newLiteral);
        }
        return base.Visit(node);
    }
}
```

```csharp

public class ConcatToTemplateLiteral : AstConverter
{
    public override Node? Visit(Node? node)
    {
        if (node is not CallExpression b)
            return base.Visit(node);

        if (b.Callee is not MemberExpression { Property: Identifier { Name: "concat" } } m) return base.Visit(node);

        var items = new List<Expression> { m.Object };
        foreach (var argument in b.Arguments)
        {
            items.Add(argument);
        }

        var quasis = new List<TemplateElement>();
        var expressions = new List<Expression>();
        var currentQuasi = "";

        void AddQuasi(string s, bool isTail)
        {
            var quasiValue = new TemplateElement.TemplateElementValue()
            {
                Raw = s
            };
            quasis.Add(new TemplateElement(quasiValue, isTail));
            currentQuasi = "";
        }

        foreach (var argument in items)
        {
            var isTail = b.Arguments.Last() == argument;
            if (argument is Literal literal)
            {
                currentQuasi += literal.GetValueString();
                if (isTail)
                {
                    AddQuasi(currentQuasi, isTail);
                }
            }
            else
            {
                AddQuasi(currentQuasi, isTail);
                expressions.Add(argument);
            }
        }

        var nodes = NodeList.Create<TemplateElement>(quasis);
        return new TemplateLiteral(in nodes, NodeList.Create(expressions));
    }
}
```
My use-case is simplifying the AST enough that I can extract strings/information from javascript files.

```csharp
    [Fact]
    public void CanCombineSimpleLiteral()
    {
        // Arrange
        var program = TestBase.ParseExpression("var x = '1' + '2';");
        var visitor = new CombineBinaryLiterals();

        // Act
        var result = visitor.Visit(program);

        // Assert
        Assert.NotSame(program, result);

        var literalNode = FindNearTypeOfDescendTyped(typeof(Literal), result);
        literalNode.As<Literal>().StringValue.Should().Be("12");
    }
```
```csharp
    [Fact]
    public void TestCombineLiterals()
    {
        // Arrange
        var program = TestBase.ParseExpression("var x = '1'.concat('2');");
        var visitor = new ConcatToTemplateLiteral();

        // Act
        var result = visitor.Visit(program);

        // Assert
        Assert.NotSame(program, result);

        var literalNode = FindNearTypeOfDescendTyped(typeof(TemplateLiteral), result);
        var quasi = literalNode.As<TemplateLiteral>().Quasis.FirstOrDefault();
        quasi.Should().NotBeNull();
        quasi.Tail.Should().BeTrue();
        quasi.Value.Raw.Should().Be("12");
    }
    
```

It's fairly clean, but being able to avoid checking the type of every single node would be ideal.

adams85:
Alright, then I suggest we merge this PR first, then I submit another PR with the proposed changes.

@Xicy Could you prepare this PR for merge (i.e. make it up-to-date with the base branch and make sure it compiles without errors)?

Xicy:
> @Xicy Could you prepare this PR for merge (i.e. make it up-to-date with the base branch and make sure it compiles without errors)?

Every thing is fine for now. Thank you for information especially Rebase. I mixed up this pr sorry for that.

lahma:
@Xicy and @adams85 thank you for your great collaboration. I don't have the subject matter expertise and I'm really happy that you can have an open conversation and improve the design. I've merged that changed and I believe this is more to come. Let's enjoy this v3 beta phase 😉

PR #248, opened 2022-05-28
Class parsing improvements

lahma:
Trying to fix as many class parsing related problems as possible, without introducing new failing test cases.

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -1,4 +1,4 @@
-using System.Diagnostics.CodeAnalysis;
+﻿using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using Esprima.Ast;
 
@@ -25,12 +25,14 @@ public partial class JavaScriptParser
             IsModule = false;
             AllowIn = true;
             AllowStrictDirective = true;
-            AllowSuper = false;
+            AllowSuperCall = false;
+            AllowSuperProperty = false;
             AllowYield = true;
-            IsAsync = false;
+            AllowAwait = false;
             FirstCoverInitializedNameError = null;
             IsAssignmentTarget = false;
             IsBindingElement = false;
+            InClassFieldInit = false;
             InFunctionBody = false;
             InIteration = false;
             InSwitch = false;
@@ -59,11 +61,13 @@ public partial class JavaScriptParser
         public bool IsModule;
         public bool AllowIn;
         public bool AllowStrictDirective;
-        public bool AllowSuper;
+        public bool AllowSuperCall;
+        public bool AllowSuperProperty;
         public bool AllowYield;
-        public bool IsAsync;
+        public bool AllowAwait;
         public bool IsAssignmentTarget;
         public bool IsBindingElement;
+        public bool InClassFieldInit;
         public bool InFunctionBody;
         public bool InIteration;
         public bool InSwitch;
@@ -569,6 +573,19 @@ public partial class JavaScriptParser
         return token.Type == TokenType.Identifier && keyword.Equals((string) token.Value!);
     }
 
+    // Return true if the next token matches the specified contextual keyword and consumes the next token.
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private bool ConsumeContextualKeyword(string keyword)
+    {
+        if (MatchContextualKeyword(keyword))
+        {
+            NextToken();
+            return true;
+        }
+        return false;
+    }
+
     // Return true if the next token is an assignment operator
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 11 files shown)


jogibear9988:
@lahma are these changes I should port back to esprima-next?

lahma:
@jogibear9988 I believe that is the case when I get this finished. My goal is to get all outstanding class-parsing problems sorted and I've already found some interesting bugs and construct problems. The end result might be this is going to be relatively big PR (if I get it finished!).

src/Esprima/JavascriptParser.cs
```diff
@@ -684,10 +701,14 @@ private protected virtual Expression ParsePrimaryExpression()
         switch (_lookahead.Type)
         {
             case TokenType.Identifier:
-                if ((_context.IsModule || _context.IsAsync) && "await".Equals(_lookahead.Value))
+                if ((_context.IsModule || _context.AllowAwait) && "await".Equals(_lookahead.Value))
                 {
                     TolerateUnexpectedToken(_lookahead);
                 }
+                if (_context.InClassFieldInit && "arguments".Equals(_lookahead.Value))
```
adams85, line 708: `_context.InClassFieldInit` is never set, it always has its default value (`false`).

src/Esprima/JavascriptParser.cs
```diff
@@ -961,34 +982,31 @@ private BlockStatement ParsePropertyMethod(ref ParsedParameters parameters, out
...
-        _context.AllowYield = previousAllowYield;
-
-        return Finalize(node, new FunctionExpression(null, NodeList.From(ref parameters.Parameters), method, isGenerator, hasStrictDirective, false));
-    }
-
-    private FunctionExpression ParsePropertyMethodAsyncFunction(bool isGenerator)
-    {
-        var node = CreateNode();
+        var previousStrict = _context.Strict;
+        _context.Strict = false;
```
adams85, line 990: This looks wrong.

For example
```js
({ m(x = 0644) { } })
```
compiles but
```js
'use strict';
({ m(x = 0644) { } })
```
doesn't.

On top of that, the entire body of classes is in strict mode explicitly IIRC.

src/Esprima/JavascriptParser.cs
```diff
@@ -1329,7 +1349,10 @@ private Node ReinterpretExpressionAsPattern(Node expr)
                 {
                     if (property is Property p)
                     {
-                        p._value = ReinterpretExpressionAsPattern(p.Value);
+                        if (p._value is not null)
+                        {
+                            p._value = ReinterpretExpressionAsPattern(p.Value);
+                        }
```
adams85, line 1355: I'm puzzled about this change. In what circumstances can a property value be `null` in binding patterns? Am I missing here something? (If so, that would have consequences for the AST model and other features like AST to JS... )

src/Esprima/JavascriptParser.cs
```diff
@@ -2395,7 +2427,17 @@ private void CheckPatternParam(ref ParsedParameters options, Node param)
...
+                    {
+                        if (p.Value is not null)
+                        {
+                            CheckPatternParam(ref options, p.Value);
+                        }
+                    }
+                    else
+                    {
+                        CheckPatternParam(ref options, property);
+                    }
```
adams85, line 2440: Same question as above.

src/Esprima/JavascriptParser.cs
```diff
@@ -4506,8 +4561,10 @@ private FunctionExpression ParseGeneratorMethod(bool isAsync = false)
 
         _context.AllowYield = true;
```
adams85, line 4562: Seems suspicious. Yield is enabled in the parameter list just because it's a generator method? Shouldn't it be enabled for the method body only?

src/Esprima/JavascriptParser.cs
```diff
@@ -4506,8 +4561,10 @@ private FunctionExpression ParseGeneratorMethod(bool isAsync = false)
...
         var parameters = ParseFormalParameters();
-        _context.AllowYield = false;
+
+        _context.AllowYield = true;
```
adams85, line 4565: This assignment may be redundant depending on the issue above.

src/Esprima/JavascriptParser.cs
```diff
@@ -4623,20 +4679,19 @@ private ArrayList<Decorator> ParseDecorators()
     {
         var decorators = new ArrayList<Decorator>();
 
-        while (Match("@"))
+        while (ConsumeMatch("@"))
```
adams85, line 4682: This way the leading `@` will be excluded from the range of the decorator. AST Explorer includes it, so I think this change should be reverted.

src/Esprima/JavascriptParser.cs
```diff
@@ -4645,175 +4700,152 @@ private ClassElement ParseClassElement(ref bool hasConstructor)
         var isAsync = false;
         var isGenerator = false;
         var isPrivate = false;
+        var constructor = false;
```
adams85, line 4703: `computed`, `isAsync` and `isGenerator` is unused.

adams85 (commented):
I rebased this PR to help you (with cleaning up the mess which partly I created 😄 ) and to see where we are with #330. I did a preliminary review to let you know of the (possible) issues I found during rebasing.

lahma:
Thanks for doing the rebase, I need to go trough this again with some thought at some point....

src/Esprima/JavascriptParser.cs
lahma: I probably need to check the (now enabled) test suite and check if the breakpoint is hit and in which conditions.

adams85:
No problem. I think it might be the best to finish the fixes that this PR currently contains first and merge it without including more. Then to address parser issues not with megaPRs but with smaller ones focusing on one bug/missing feature/etc so we have less merge conflicts or at least they would be more easy to handle them.

jogibear9988:
I'd also like if thepull requests are small, so it's easier for me to see if I need the change also on esprima-next.
I woule be happy if the projects could stay in sync a little bit.
So we should check after this is merged to include my fixes I mentioned here: 
https://github.com/sebastienros/esprima-dotnet/issues/330#issuecomment-1239963406

lahma:
Closing as obsolete.

PR #249, opened 2022-05-28
Update GitHubActionsTestLogger to latest preview

lahma:
This will reduce summary size which seems to be a problem with Esprima-size output.


lahma:
OK, now the summary looks decent and performant: https://github.com/sebastienros/esprima-dotnet/actions/runs/2400457387

PR #252, opened 2022-06-01
fixed TemplateLiteral ChildNodes order

### LINKED ISSUE #177: TemplateLiteral orders its child nodes with TemplateElements first and tokens last
Take this JS for example:

    var name = "Bob";
    var age = 30;
    var s = `Name: ${name}, Age: ${age}`;

The child nodes are ordered like this:

image

I had to write this code as a workaround which basically iterates through the TemplateElements and then looks ahead in the collection for its corresponding token. It works, but it seems like the nodes should be ordered as they appear in the string so that you can just do a simple `for each` through them.

        public static IEnumerable<Node> EnumerateChildNodes(this TemplateLiteral node)
        {
            var max = (int)Math.Ceiling((double)node.ChildNodes.Count / 2);

            for (int i = 0; i < max - 1; i++)
            {
                yield return node.ChildNodes[i];

                yield return node.ChildNodes[i + max];
            }

            yield return node.ChildNodes[max - 1];
        }

Xicy:
Applied solution for ChildNodes order according to issue #177 

fixes #177

### DIFF
--- a/src/Esprima/Ast/TemplateLiteral.cs
+++ b/src/Esprima/Ast/TemplateLiteral.cs
@@ -19,11 +19,22 @@ namespace Esprima.Ast
         public ref readonly NodeList<TemplateElement> Quasis => ref _quasis;
         public ref readonly NodeList<Expression> Expressions => ref _expressions;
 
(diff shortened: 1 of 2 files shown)


lahma (approved):
Thanks!

PR #257, opened 2022-06-04
Update AstJson

adams85:
It would be nice to update `AstJson` before releasing the new major version because it's still somewhat bit rotten (missing overrides, missing handling of a few properties). I didn't touch this as part of https://github.com/sebastienros/esprima-dotnet/pull/253 because it would result in a ton of failed tests. Any idea how to approach this?

### DIFF
--- a/src/Esprima/Utils/AstJson.cs
+++ b/src/Esprima/Utils/AstJson.cs
@@ -347,6 +347,11 @@ public class AstToJsonConverter : AstJson.IConverter
             {
                 Member("body", program.Body, e => (Node) e);
                 Member("sourceType", program.SourceType);
+
+                if (program is Script s)
+                {
+                    Member("strict", s.Strict);
+                }
             }
 
             return program;
@@ -377,6 +382,7 @@ public class AstToJsonConverter : AstJson.IConverter
                 Member("body", functionDeclaration.Body);
                 Member("generator", functionDeclaration.Generator);
                 Member("expression", functionDeclaration.Expression);
+                Member("strict", functionDeclaration.Strict);
                 Member("async", functionDeclaration.Async);
             }
 
@@ -576,6 +582,7 @@ public class AstToJsonConverter : AstJson.IConverter
                 Member("body", arrowFunctionExpression.Body);
                 Member("generator", arrowFunctionExpression.Generator);
                 Member("expression", arrowFunctionExpression.Expression);
+                Member("strict", arrowFunctionExpression.Strict);
                 Member("async", arrowFunctionExpression.Async);
             }
 
@@ -705,6 +712,16 @@ public class AstToJsonConverter : AstJson.IConverter
             return identifier;
         }
 
+        protected internal override object? VisitPrivateIdentifier(PrivateIdentifier privateIdentifier)
+        {
+            using (StartNodeObject(privateIdentifier))
+            {
+                Member("name", privateIdentifier.Name);
+            }
+
+            return privateIdentifier;
+        }
+
         protected internal override object? VisitFunctionExpression(FunctionExpression functionExpression)
         {
             using (StartNodeObject(functionExpression))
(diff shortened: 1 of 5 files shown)


lahma:
I don't have strong opinions, except that I would like that the output still is easily diffable against a reference implementation. I generally use https://astexplorer.net/ , but in the future might be nice automation to run, say acorn, and then diff the output against each file. The current set is quite limited. One idea would also to change to snapshot testing, probably using https://github.com/VerifyTests/Verify .

Would the failed tests be because of the current snapshots? Would it make sense to approach this with changing to using Verify and then get easier workflow for snapshotting and updating blessed output (PR has diffs)?

adams85:
> Would the failed tests be because of the current snapshots?

As far as I can see, that's the case. The updated `AstJson` includes some information in the actual json which is not present in the expected (reference) json files (for example, see [this change](https://github.com/sebastienros/esprima-dotnet/pull/257/commits/148d8bdfcec5e5530d12387986d0d1c7e5590dc2#diff-ba2530ce3457ecc73a5e6d03b02b1c2511a058af21c66470a813e788b5fab368R330-R333)). There were also changes in the AST (e.g. class properties are represented now as `PropertyDefinition`), so I assume these changes also make some tests fail.

Fixing the snapshots manually by adding the missing bits is hopeless as we're talking about 1000 failing tests here.

As I see our options are the following:

1. We can probably regenerate the snapshots. It looks like there's a switch [here](https://github.com/adams85/esprima-dotnet/blob/148d8bdfcec5e5530d12387986d0d1c7e5590dc2/test/Esprima.Tests/Fixtures.cs#L12) for this purpose. I'm just not sure it's a correct procedure to simply regenerate the snapshots and call it a day.

2. "Approach this with changing to using Verify". At first glance, Verify would be more or less equivalent to what we do now by converting to json and comparing with a snapshot. However, it looks pretty tedious to convert our tests to Verify. I'd avoid this unless there's a clear advantage.

3. If `AstJson` is only meant for testing purposes, we might as well just kick the can down the road. We may remove `AstJson` from the public API, move it to the test project and leave it alone for now.

I can't spare much time for this, so I'm leaning toward option 1. However, there are 57 failing tests now (without applying this PR). Option 1 could be done safely only if there were no failing tests because it's not obvious if the aforementioned tests are false negatives or actual issues. Tracking down these failing tests would also take a lot of time (at least, for me).

lahma:
I trust your judgment here, so we can go with the first option. This is going to be incremental and we are I no hurry, betas and previews can be published and Jint can use snapshots.

Snapshot testing has the advantage of opening diff viewer instead of the horrible Assert experience with string comparison with some 10 characters  but we can improve later.

adams85:
Alright, then option 1 it is. However, I'll leave those 57 tests failing currently unchanged. That'll be a problem to solve in the future.

adams85:
I think in the meantime I could come up with an even better idea:

As I've learned the reference JSON files (which contain the expected ASTs for the test fixtures) stem from the original Esprima project. Our `AstJson` class is used for converting the actual ASTs into JSON files which are then compared to the reference JSONs in the assert phase.

As there are minor differences between the AST produced by our parser compared to the AST produced by original Esprima, these won't match as expected.

This problem can be resolved in two ways:
1. By modifying the original reference JSONs to match our output. As far as I can see, this is what's been done so far.
2. By modifying our AST->JSON conversion so that its output matches the original reference JSONs.

I think the second option would be a much better approach. So I've [introduced a switch in `AstJson.Options`](https://github.com/sebastienros/esprima-dotnet/pull/257/commits/9c45a830d3ae291afcc732532cb80f1f946b4d88#diff-ba2530ce3457ecc73a5e6d03b02b1c2511a058af21c66470a813e788b5fab368R33) which makes this possible: by enabling this, we can produce output in a format suitable for testing purposes. Not only can we avoid constantly adjusting our reference files in this way, but end users are also allowed to generate an exact JSON representation of the AST produced by *our* parser.

However, I only partially implemented this "compatibility mode": the output now matches *our current* reference JSONs. But these are not identical to the reference JSONs of original Esprima. That is, we're somewhere between option 1 and 2 currently.

But reaching option 2 is another story, copying over the original JSONs and modify our AST->JSON conversion to match these is another item for the backlog. What I intended to achieve with this PR is complete now.

lahma (approved):

lahma:
Thanks!

jogibear9988:
I don't know if we need to match the original esprima ones. There seems absolutely no development happening there. SO I do not expect them to change.
And I think the jsons in my fork already do not match the esprima ones any more.

PR #258, opened 2022-06-04
Refactor JSX parsing into an isolated extension on top of standard JS parsing

adams85:
This PR aims to separate JSX parsing (+ related data structures and logic) from standard JavaScript parsing. It introduces a simple extension model and implements JSX-related logic using that.

Key changes:
* Everything JSX-related has been moved into the `Esprima.Jsx` namespace.
* `JavaScriptParser` does only standard JS parsing now. For JSX parsing we provide the `JsxParser` class. This is inherited from `JavaScriptParser` just as [it's done in original Esprima](https://github.com/jquery/esprima/blob/512cd66c6ffd6083144b0150f09670e426252776/src/jsx-parser.ts#L60).
* A new node type named `Extension` has been added to the `Nodes` enumeration. (The idea comes from .NET expression trees which have the same concept). Non-standard nodes (incl. JSX nodes) should use this as node type. In the case of JSX nodes, a new enumeration named `JsxNodeType` has been introduced for describing the actual node type ([see](https://github.com/sebastienros/esprima-dotnet/pull/258/commits/195dda2b5918a9201caff4548cfa4d0f7c253260#diff-fb532590119b5e5d5345b1d0ebddf2a0f6f9bfeff40e4520efe59298493f1fc6R10-R15)).
* The base visitor (`AstVisitor`) throws when it encounters an extension node. (This behavior [can be overridden]( https://github.com/sebastienros/esprima-dotnet/pull/258/commits/a0a7ac194462ecd7349158851741612a197ac6ce#diff-4d5f076f726de090debe3f7afe346e7a2bbff7ea4343797e33021c0b95f534eeR712) though.) In addition a new interface named `IJsxAstVisitor` has been introduced and by implementing this, existing visitors can easily be made capable of handling JSX nodes ([see also](https://github.com/sebastienros/esprima-dotnet/pull/258/commits/a0a7ac194462ecd7349158851741612a197ac6ce#diff-fb532590119b5e5d5345b1d0ebddf2a0f6f9bfeff40e4520efe59298493f1fc6R23)).
* As a matter of fact, this is already done for all the visitors provided by the library. All plain visitors have their JSX-enabled counterparts:

  * `AstVisitor` -> `JsxAstVisitor`
  * `AstRewriter` -> `JsxAstRewriter`
  * `AstVisitorEventSource` -> `JsxAstVisitorEventSource`
  * Also `AstJson` has been updated. The AST->JSON conversion logic has been refactored into the `AstToJsonConverter` class, and the corresponding `JsxAstToJsonConverter` has also been added.

@Xicy As the original implementer of the JSX feature, what's your opinion on the proposed changes? Are they acceptable to you? The internals have gotten a bit more complicated, especially JSX visitation which has a multiple inheritance issue. Due to this, I had to do some gymnastics as it's not supported in C# directly. The public APIs remained roughly the same. You just need to inherit from the JSX-enabled visitors when you deal with an AST containing JSX nodes. The rest should work as before.


Xicy:
That is a great work 🎉

Actually I loved the Extensions approach. I will deep look to tomorrow. 

Just I want remark something. If we try to protect same result of Esprima, we don't have to change value of the "type" inside json data.

adams85:
> Just I want remark something. If we try to protect same result of Esprima, we don't have to change value of the "type" inside json data.

Oh, I see now where you're coming from. I wasn't aware that the test fixtures are borrowed from the original Esprima projects. Thus, of course, it makes perfect sense to stay consistent with its naming. I've corrected this.

Thanks for pointing this out and I'm looking forward to your further remarks!

src/Esprima/Jsx/JsxToken.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Esprima.Jsx
+{
+    public enum JsxTokenType
+    {
+        JsxIdentifier,
+        JsxText
+    };
+
+    public sealed class JsxToken : Token
```
lahma, line 15: this might not fly well with my attempt of changing Token to struct... but I think I can work around it struct-based approach seems beneficial

lahma (approved):

Xicy:
Looks good to me. Everything works fine 🎉

src/Esprima/Jsx/JsxToken.cs
adams85: That also seems beneficial to me if you can pass `Token` around without copying (i.e. by-ref).

As for the problem of inheritance, you may add an `object` field named something like `ExtensionData` to `Token`. Then `JsxParser` could store the JSX token type in that field. This info is accessed only [at one place in `JsxParser`] (https://github.com/sebastienros/esprima-dotnet/pull/258/commits/ae86c4e53e349ddac2cb9889c8eab45db84544b0#diff-3369e40bf400ba131a958de181b5592a878f88e30027aa3a1a8c2dae81e5c042R636) anyway.

adams85:
On a second thought, I'm not sure about the namespace structure any more. Repeating the whole structure under the `Jsx` namespace feels a bit weird. Wouldn't be it better the other way around, that is, e.g. `Esprima.Ast.Jsx` instead of `Esprima.Jsx.Ast`?

@lahma Still not convinced of putting JSX into another package? Of course, assembly sizes are not a concern in .NET land usually (apart from Blazor-like use cases) but being an optional thing, it looks like a natural choice to me.

lahma:
> On a second thought, I'm not sure about the namespace structure any more. Repeating the whole structure under the `Jsx` namespace feels a bit weird. Wouldn't be it better the other way around, that is, e.g. `Esprima.Ast.Jsx` instead of `Esprima.Jsx.Ast`?
> 
> @lahma Still not convinced of putting JSX into another package? Of course, assembly sizes are not a concern in .NET land usually (apart from Blazor-like use cases) but being an optional thing, it looks like a natural choice to me. 

I think this is for @sebastienros to decide. Requires new NuGet handling etc.

adams85:
Ok, let's finish this and deal with the question of the separate package in another PR if we're given the green light.

I've updated this one and changed the namespace structure, too. Maybe it's somewhat saner this way.

So I think it's ready for merge.

lahma:
Great, thanks for going trough this iteration!

PR #263, opened 2022-06-07
Minor corrections to AST

adams85:
This PR addresses a few anomalies of the AST.

1. Nullability of `Import.Source` seems wrong. How to import something if one doesn't specify what? Also according to [the spec](https://tc39.es/ecma262/#sec-import-call-runtime-semantics-evaluation) it's not optional. Am I missing something here? (BTW, the spec also says that this is an `AssignmentExpression` but we have declared it as a more general `Expression`. But looking at the parser code, I'm not 100% sure that it's always an `AssignmentExpression` in our implementation. Is it safe to change this to `AssignmentExpression`?)
2. `PropertyDefinition.Kind` was not set.
3. `PropertyDefinition.Value` can be null. Try to parse this or view it in AST explorer:
   ```js
   class Rectangle {
     height = 0;
     width;
     constructor(height, width) {
       this.height = height;
       this.width = width;
     }
   }
   ```

To remedy the last one, I had to use that ugly trick which I did in the case of `ImportDeclarationSpecifier` earlier (IIRC). And while doing this, I again encountered the downside of using fields instead of properties in the AST classes:

There are some fields which can't be declared as read-only because they are re-assigned during parsing. Most of the AST classes are immutable but not all of them. This is not an ideal situation. Unfortunately, without some unholy hack we can't make all fields read-only, so properties look like the only way out.

We may do some refactoring as follows:

```csharp
public sealed class AssignmentPattern : Expression
{
    //public readonly Expression Left;
    //public Expression Right;

    public Expression Left { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
    public Expression Right { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; internal set; }

    // ...
}
```

But we can just hope that there will be no significant performance degradation, we will only know after refactoring and doing benchmarks.

If you say so, I can take the task of refactoring the fields into properties.

### DIFF
--- a/src/Esprima/Ast/ClassElement.cs
+++ b/src/Esprima/Ast/ClassElement.cs
@@ -1,17 +1,20 @@
-﻿#nullable disable
-
-namespace Esprima.Ast
+﻿namespace Esprima.Ast
 {
     public abstract class ClassProperty : Expression
     {
-        public PropertyKind Kind;
+        public readonly PropertyKind Kind;
 
-        public Expression Key; // Identifier, Literal, '[' Expression ']'
-        public bool Computed;
-        public Expression Value;
+        public readonly Expression Key; // Identifier, Literal, '[' Expression ']'
+        public readonly bool Computed;
 
-        protected ClassProperty(Nodes type) : base(type)
+        public Expression? Value => GetValue();
+        protected abstract Expression? GetValue();
+
+        protected ClassProperty(Nodes type, PropertyKind kind, Expression key, bool computed) : base(type)
         {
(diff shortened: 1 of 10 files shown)


lahma (approved):
I did similar property changes in my class parsing improvements branch, might as well get into main before that, who doesn't like merge conflicts 😝 

Changing to getters, I think it would sufficient to run `AstTreeWalkBenchmark ` and check how/if results change. It should go trough large-enough JS file and visit everything. If change is all about field -> properties, the benchmark should show the penalty.

PR #268, opened 2022-06-08
Improve test compatibility 

adams85:
This PR finishes the work started in https://github.com/sebastienros/esprima-dotnet/pull/257. Motivation is explained [here](https://github.com/sebastienros/esprima-dotnet/pull/257#issuecomment-1148006877).

With these changes, we can reuse the test fixtures of the original Esprima project *without modification* as our AST->JSON conversion (see `AstToJsonConverter`) is now also able to transform our AST into the expected format of the original fixtures. This compatibility mode can be enabled in the newly introduced `fixtures-metadata.json` (by means of which further additional information is provided for the conversion to produce the desired output).

Those fixtures which are added by us doesn't use the compatibility mode. These have been updated to reflect the current state of our version of the AST.

Remark: you may have to delete the `test\Esprima.Tests\Fixtures` folder and do a `git checkout .` in your working copy to get the correct line endings in the fixture JS files. Otherwise, you can get mysteriously failing tests because location and range data is sensitive to the line endings used. (BTW, this is handled by means of `test\Esprima.Tests\Fixtures\.gitattributes`).


adams85:
Oh, and one further remark about the remark:

There's a pretty [suspicious test fixture](https://github.com/adams85/esprima-dotnet/blob/4ee3fb30e19170887b7c572d77f91e49086831f9/test/Esprima.Tests/Fixtures/expression/primary/literal/string/migrated_0017.js) which looks like a false positive. The test passes in the case `\r\n` line endings but fails in the case of `\n`. Can someone more familiar with the parser tell if this is the intended behavior?

adams85:
As far as I'm concerned, it's ready.

> Can't remember the history of this. But there are some test cases that meant specifically to be run on windows (needs the gitattributes set).

Yeah, that's correct, I revised `.gitattributes`. However, the test fixture in question looks like a false positive, i.e., it passes although there may be a bug. It's just covered by  `.gitattributes`. I'll open an issue for this.

adams85:
> However, the test fixture in question looks like a false positive, i.e., it passes although there may be a bug. It's just covered by .gitattributes. I'll open an issue for this.

I've looked into this and it turned out that we're ok, it's not a bug. Actually, this looks like a problem on original Esprima's side. They make assertions with CRLF but [their `.gitattributes`](https://github.com/jquery/esprima/blob/512cd66c6ffd6083144b0150f09670e426252776/test/.gitattributes) doesn't reflect this.

adams85:
You're welcome!

Issue #264, opened 2022-06-08
Change fields to properties


lahma:
As it's been a nagging pain, we should try what perf consequences we get from doing so. Should have large JS file benchmark with no-op sealed `AstVisitor` running and see if there's anything of concern.

I'm creating some issues to track v3 ideas and progress.

/cc @adams85

adams85:
This is also related to https://github.com/sebastienros/esprima-dotnet/issues/66

adams85:
Can be closed (resolved by https://github.com/sebastienros/esprima-dotnet/pull/270)

PR #270, opened 2022-06-10
Change fields to properties in AST public API

adams85:
This PR addresses https://github.com/sebastienros/esprima-dotnet/issues/264 and, partially, https://github.com/sebastienros/esprima-dotnet/issues/66

Phew, this was some epic sh*t. It took all my text editing skills... 😄 But according to the benchmark it was worth it:

Before:

|  Method |        Job |              Runtime |    Toolchain |     Mean |    Error |   StdDev | Ratio | Rank | Allocated |
|-------- |----------- |--------------------- |------------- |---------:|---------:|---------:|------:|-----:|----------:|
|   Visit | Job-ZDQNTH |             .NET 6.0 |        net60 | 15.18 ms | 0.031 ms | 0.026 ms |  0.97 |    1 |      16 B |
|   Visit | Job-IEZQPH |        .NET Core 3.1 | netcoreapp31 | 15.24 ms | 0.029 ms | 0.027 ms |  0.98 |    1 |         - |
|   Visit | Job-WZKQTX | .NET Framework 4.6.2 |       net462 | 15.63 ms | 0.014 ms | 0.012 ms |  1.00 |    2 |         - |
|         |            |                      |              |          |          |          |       |      |           |
| Rewrite | Job-ZDQNTH |             .NET 6.0 |        net60 | 37.70 ms | 0.134 ms | 0.125 ms |  0.75 |    1 |      75 B |
| Rewrite | Job-IEZQPH |        .NET Core 3.1 | netcoreapp31 | 40.52 ms | 0.061 ms | 0.054 ms |  0.81 |    2 |         - |
| Rewrite | Job-WZKQTX | .NET Framework 4.6.2 |       net462 | 50.06 ms | 0.298 ms | 0.279 ms |  1.00 |    3 |         - |

After:

|  Method |        Job |              Runtime |    Toolchain |     Mean |    Error |   StdDev | Ratio | Rank | Allocated |
|-------- |----------- |--------------------- |------------- |---------:|---------:|---------:|------:|-----:|----------:|
|   Visit | Job-PALKNJ |             .NET 6.0 |        net60 | 15.26 ms | 0.046 ms | 0.043 ms |  0.98 |    2 |      16 B |
|   Visit | Job-SJJCSG |        .NET Core 3.1 | netcoreapp31 | 14.96 ms | 0.019 ms | 0.017 ms |  0.96 |    1 |         - |
|   Visit | Job-QAMQPG | .NET Framework 4.6.2 |       net462 | 15.55 ms | 0.016 ms | 0.013 ms |  1.00 |    3 |         - |
|         |            |                      |              |          |          |          |       |      |           |
| Rewrite | Job-PALKNJ |             .NET 6.0 |        net60 | 36.63 ms | 0.190 ms | 0.178 ms |  0.72 |    1 |      75 B |
| Rewrite | Job-SJJCSG |        .NET Core 3.1 | netcoreapp31 | 43.10 ms | 0.082 ms | 0.073 ms |  0.85 |    2 |         - |
| Rewrite | Job-QAMQPG | .NET Framework 4.6.2 |       net462 | 50.52 ms | 0.119 ms | 0.111 ms |  1.00 |    3 |         - |

So it looks like aggressive inlining did its job. (I didn't measure parsing because construction of the tree remained the same as I kept the fields for the constructors and other internal parts.) However, to be 100% certain, it'd be great if someone repeated the benchmarks and verified these result.

With these changes, the AST is now almost fully immutable. It's almost because I haven't touched `Node.Location` and `Node.Range`. These are still public fields and I'm somewhat puzzled as to what to do with them. The crux of the problem is that these properties aren't included in the constructors but should be settable also by end users. Including them in the ctors would be a massive BC, so that's absolutely out of question. Our options are the following:
1. We can simply wrap them in properties. E.g.
    ```csharp
    public Range Range { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; [MethodImpl(MethodImplOptions.AggressiveInlining)] init; }
    ```
    However, this is certainly worse performance-wise than we have currently (i.e. simple fields) because `Range` and `Location` are value types, so there would be inevitable copying.
2. We can return by-ref but then the properties can't have a setter. We could provide some `SetRange` and `SetLocation` methods but then it's not immutable again. For init-only set we need a property. Thus, if we want by-ref return we'd need two properties:
    ```csharp
    public ref readonly Range RangeRef { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => ref _range; }
    public Range Range { get => _range; init => _range = value; }
    ```
    This is ugly but the only way I'm aware of to retain perf. and make the API immutable for consumers at the same time.
3. We leave them alone and have a semi-immutable API.

Which one do you prefer?

### DIFF
--- a/src/Esprima/Ast/AccessorProperty.cs
+++ b/src/Esprima/Ast/AccessorProperty.cs
@@ -1,17 +1,11 @@
+using System.Runtime.CompilerServices;
 using Esprima.Utils;
 
 namespace Esprima.Ast;
 
 public sealed class AccessorProperty : Node
 {
-    /// <summary>
-    /// <see cref="Expression" /> | <see cref="PrivateIdentifier" />
-    /// </summary>
-    public readonly Expression Key;
-    public readonly Expression? Value;
-    public readonly bool Computed;
-    public readonly bool Static;
-    public readonly NodeList<Decorator> Decorators;
+    private readonly NodeList<Decorator> _decorators;
 
     public AccessorProperty(Expression key, Expression? value, bool computed, bool isStatic, in NodeList<Decorator> decorators) : base(Nodes.AccessorProperty)
     {
@@ -19,8 +13,17 @@ public sealed class AccessorProperty : Node
         Value = value;
         Computed = computed;
         Static = isStatic;
-        Decorators = decorators;
+        _decorators = decorators;
     }
+
+    /// <remarks>
+    /// <see cref="Expression" /> | <see cref="PrivateIdentifier" />
+    /// </remarks>
+    public Expression Key { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public Expression? Value { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public bool Computed { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public bool Static { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public ref readonly NodeList<Decorator> Decorators { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => ref _decorators; }
 
     public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(NodeList.Create(CreateChildNodes()));
 
--- a/samples/Esprima.Benchmark/Esprima.Benchmark.csproj
+++ b/samples/Esprima.Benchmark/Esprima.Benchmark.csproj
@@ -3,7 +3,8 @@
     <AssemblyName>Esprima.Benchmark</AssemblyName>
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net462;netcoreapp3.1;net6.0</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <None Include="..\..\test\Esprima.Tests\Fixtures\3rdparty\**" CopyToOutputDirectory="PreserveNewest" LinkBase="3rdparty" />
--- a/samples/Esprima.Benchmark/VisitorBenchmark.cs
+++ b/samples/Esprima.Benchmark/VisitorBenchmark.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Jobs;
+using Esprima.Utils;
+
+namespace Esprima.Benchmark
+{
+    [MemoryDiagnoser]
+    [RankColumn]
+    [CsvExporter]
+    [MarkdownExporterAttribute.GitHub]
+    public class VisitorBenchmark
+    {
+        private const string FileName = "bundle";
+
+        private Ast.Program _ast;
+        private AstVisitor _visitor;
+        private AstRewriter _rewriter;
+
+        [GlobalSetup]
+        public void Setup()
+        {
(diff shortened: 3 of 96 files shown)


lahma:
I can confirm that there is no noticeable difference, run the results for main and this using net6.0:

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=7.0.100-preview.2.22153.17
  [Host]     : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT
  DefaultJob : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT


```

| Old |Visit|12.61 ms|0.124 ms|20 B|
| **New** |	| **12.38 ms (-2%)** | **0.069 ms** | **20 B (0%)** |
| Old |Rewrite|23.20 ms|0.186 ms|33 B|
| **New** |	| **22.64 ms (-2%)** | **0.257 ms** | **33 B (0%)** |


> Which one do you prefer?

Copying `Range` *shouldn't* be a big problem as it's a small struct. `Location` in the other hand is somewhat bigger already and might cause some perf hit. Maybe something that needs to be measured again. This time I think the benchmark would be parsing speed though.

We can leave that to another PR/later stage too though. So option 1 or 3 I guess.

src/Esprima/Ast/AccessorProperty.cs
```diff
@@ -1,27 +1,34 @@
...
+        _value = value;
+        _computed = computed;
+        _static = isStatic;
+        _decorators = decorators;
     }
 
+    /// <remarks>
+    /// <see cref="Expression" /> | <see cref="PrivateIdentifier" />
+    /// </remarks>
+    public Expression Key { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => _key; }
```
lahma, line 26: For a shorter syntax without need for backing field you could also try:

```csharp
public Expression Key { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
```
It didn't change benchmark results and resharper/rider allows you to do a mass-change with alt-enter and selecting suitable scope, like folder.

lahma (approved):
Just a note that backing field might not be needed and still get the inlining.

src/Esprima/Ast/AccessorProperty.cs
adams85: You're right, I've changed this.

adams85:
> Copying Range shouldn't be a big problem as it's a small struct. Location in the other hand is somewhat bigger already and might cause some perf hit. Maybe something that needs to be measured again. This time I think the benchmark would be parsing speed though.

> We can leave that to another PR/later stage too though. So option 1 or 3 I guess.

Then I propose option 1 with a twist: let's change the fields to internal so we can use those internally for the sake of max. perf and provide the "slow" (copying) properties for end users. See the 2. commit.

If this is acceptable than the PR is ready.

lahma:
Excellent work once again!

lahma:
Noticed one small problem with latest code on Jint side:

https://github.com/sebastienros/jint/blob/3bc93cf385795ad06af98376a99b8ed17cd2b615/Jint/Native/Json/JsonParser.cs#L522-L541

Probably needs investigation why this assignment is done or needed, I'm not familiar with this part.

adams85:
> Excellent work once again!

Thank you!

> Noticed one small problem with latest code on Jint side:

Maybe add an `InternalsVisibleTo` for Jint? Or just scrap the idea of init-only properties for `Range` and `Location` and revert them to public fields?

lahma:
InternalsVisible could work. Would be nice to know what other potentially breaking consumers there are, ones wanting to update this info. But as said, not a critical thing from Jint's side at the moment.

PR #271, opened 2022-06-10
Fix Node.ChildNodes inconsistencies

adams85:
During the overhaul of AST, I noticed that `Node.ChildNodes` implementations are pretty inconsistent. This PR addresses this (at least, to some extent).

Besides some missing bits, the core of the issue is that the `ChildNodes` enumeration may contain `null` values in the current implementation: optional nodes are returned even if they are not available (see e.g. `ReturnStatement.Argument`) and the `Elements` node list of `ArrayExpression` and `ArrayPattern` may contain `null` values as well. As I see, Jint is already prepared for this but we should inform other consumers about this. Thus, I revised the nullability annotations of `NodeList<T>` and `NodeCollection`.

(TBH, it'd be better if null values were not returned at all but that change would be a major pain which I don't have the time for ATM.)


lahma (approved):
Nice cleanup!

PR #272, opened 2022-06-11
Fix incorrect range and location of dynamic import expressions

adams85:
As the AST and visitor APIs are taking shape, I've started dogfooding the recent changes in my project. According to my tests everything works nicely except for one thing:

The range and location of dynamic import expressions are off in cases where they are contained in a statement which is terminated by `;`. It looks like a bug because earlier versions (which used `CallExpression` for representing dynamic imports) gave correct results.

I've made an attempt to fix the issue.


lahma (approved):
Output looks legit when compared what AST explorer is giving us.

PR #274, opened 2022-06-13
Sort VisitXXX methods alphabetically

adams85:
Due to the large number of AST nodes, I suggest maintaining an alphabetic order of `VisitXXX` methods in our visitors. This makes the code easier to navigate and work with.

This is a code quality improvement, no logic or structure modified. Unfortunately, it kind of messes up git history of the files involved but I think this is a right time to set this straight.

### DIFF
--- a/src/Esprima/Utils/AstJson.cs
+++ b/src/Esprima/Utils/AstJson.cs
@@ -352,241 +352,24 @@ public class AstToJsonConverter : AstJson.IConverter
             }
         }
 
-        protected internal override object? VisitProgram(Program program)
+        protected internal override object? VisitArrayExpression(ArrayExpression arrayExpression)
         {
-            using (StartNodeObject(program))
+            using (StartNodeObject(arrayExpression))
             {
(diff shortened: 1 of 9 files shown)


lahma:
Would it make sense to change to file-scoped namespaces for these files too if "All is changed" anyway? Just a thought.

lahma (approved):
Ah sorry, I was on mobile and both my memory and eyes probably played tricks on me.

adams85:
No problem. BTW, the code base is still pretty mixed in this regard. File-scoped namespaces [can be enforced using .editorconfig](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#namespace-options).

adams85:
However, this PR probably concludes my contribution to v3.0. I think we've managed to sort out all the major pain points (at least, which I identified). As we discussed, there are still some rough edges here and there but nothing showstopper.

I've enjoyed working with you very much! Keep up the good work and, of course, I'll be around if you have any questions regarding the areas I worked on.

Issue #275, opened 2022-06-19
NodeCollection constructors not public


CharlieEriksen:
Hey,

I'm trying to rewrite a tree, and optimize it into higher-level "meta-nodes". So I'm creating expression types like this:

```csharp
public class HttpCallExpression : Expression
{
    public static HttpCallExpression ? Parse(CallExpression callExpr)
    {
        if (callExpr.Callee is not MemberExpression { Object: Identifier { Name: "$" }, Property: Identifier { Name: "ajax" } } ajax ||
            callExpr.Arguments[0] is not ObjectExpression arguments) return null;

        var method = arguments.Properties.GetPropertyByName("method");
        return new HttpCallExpression (callExpr, method, url, headers, data);
    }

    public HttpCallExpression (Expression node, Expression? httpMethod) : base(Nodes.ExpressionStatement)
    {

    }

    public override NodeCollection ChildNodes => new(Node);

    protected override Node Accept(AstVisitor visitor)
    {
        return visitor.VisitCallExpression(this as CallExpression);
    }
}
```

However in implementing an `Expression`, I'm running into a few issues:

- When declaring `ChildNodes`, I need to create a new NodeCollection. But none of the constructors are visible outside the project, and I can't find any helper methods for creating one. Am I missing something, or is there a reason for them not being public?
- In AstVisitor, the visit functions all seem to be marked as protected internal. What's the ideal pattern for extending the visitor to them? Is there any reason why extension methods shouldn't work for that? :) 

Thanks!
Charlie

lahma:
Did you try v3-beta2 and inheriting from `AstRewriter`? /cc @adams85

adams85:
I also recommend considering the new `AstRewriter`, which was designed to support use cases like yours.

However, regardless of this, your remarks are still valid and point out some shortcomings of the current design in respect of extensibility.

* `ChildNodes` is a [known rough edge](https://github.com/sebastienros/esprima-dotnet/pull/273#issuecomment-1153541117), it'd need some polish anyway. I may give it a shot if I can find a few free hours. Until then, you can probably just ignore it if your code doesn't need it. (The built-in visitors don't use it.) So, as a workaround, you can just throw a `NotSupportedException` for now.
[...]

PR #279, opened 2022-06-21
Add context parameter to AstVisitor.Visit

adams85:
A possible solution to the problem of maintaining contextual information during visitation. See also [this discussion](https://github.com/sebastienros/esprima-dotnet/issues/275#issuecomment-1160704010).

### DIFF
--- a/src/Esprima/Ast/ArrayExpression.cs
+++ b/src/Esprima/Ast/ArrayExpression.cs
@@ -16,9 +16,9 @@ namespace Esprima.Ast
 
         public override NodeCollection ChildNodes => GenericChildNodeYield.Yield(Elements);
 
-        protected internal override object? Accept(AstVisitor visitor)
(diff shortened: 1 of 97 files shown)


adams85:
As discussed [here](https://github.com/sebastienros/esprima-dotnet/issues/275#issuecomment-1162929943), this change wouldn't bring real value to the table because contextual information for visitations can also be maintained as an internal state of the visitor.

The pattern of saving the visitor's internal state (preferably, to the stack) before visiting some node, then restoring the state after the node has been visited is equivalent to what context parameters would achieve. A few examples:
* [tracking the parent node](https://github.com/sebastienros/esprima-dotnet/issues/275#issuecomment-1162967134)
* [eliminating `Stack<T>` usage](https://github.com/adams85/bundling/commit/370855a740615030f8c1aba36fd63e4ee8bb2656)

PR #282, opened 2022-06-26
Add a fallback to binary operator enum parser for rewriting

CharlieEriksen:
Consider an ast rewriter that attempts to rewrite a binary expression like this:
```csharp
        if (node is BinaryExpression be)
        {
            var leftNew = be.Left;

            var rightNew = be.Right;
            if (be.Left is Identifier idToResolveLeft) leftNew = ResolveFromCurrentScope(idToResolveLeft) ?? leftNew;

            if (be.Right is Identifier idToResolveRight) rightNew = ResolveFromCurrentScope(idToResolveRight) ?? rightNew;

            if (be.Left != leftNew || be.Right != rightNew)
                return new BinaryExpression(be.Operator.ToString(), leftNew, rightNew) { Location = be.Location };
        }

```

Currently, that won't work as the `ParseBinaryOperator` function assumes it's the string from the tokens. It will not accept a string like `Plus`, which you'd get from `be.Operator.ToString()`. This makes a binary expression impossible to rewrite cleanly.

This CL changes it so that the `ParseBinaryOperator` will fall back to trying to parse the op string as the enum before throwing.

I couldn't find an appropriate place to add in a test, as it doesn't seem like there are unit tests for this kind of thing. I could create a new `BinaryExpressionTests.cs`, but not sure if that'd be appropriate. Do let me know, and I'll add it in.

### DIFF
--- a/src/Esprima/Ast/BinaryExpression.cs
+++ b/src/Esprima/Ast/BinaryExpression.cs
@@ -77,7 +77,7 @@ namespace Esprima.Ast
                 "||" => BinaryOperator.LogicalOr,
                 "**" => BinaryOperator.Exponentiation,
                 "??" => BinaryOperator.NullishCoalescing,
-                _ => ThrowArgumentOutOfRangeException<BinaryOperator>(nameof(op), "Invalid binary operator: " + op)
+                _ => Enum.TryParse<BinaryOperator>(op, true, out var bo) ? bo : ThrowArgumentOutOfRangeException<BinaryOperator>(nameof(op), "Invalid binary operator: " + op)
             };
         }
 


adams85:
There's [another `BinaryExpression` ctor](https://github.com/CharlieEriksen/esprima-dotnet/blob/2f1cc74a748579638d31c344cbee80228a36e8ff/src/Esprima/Ast/BinaryExpression.cs#L43) which accepts `BinaryOperator` directly, so `new BinaryExpression(be.Operator, leftNew, rightNew) { Location = be.Location }` should be possible. Which means this change seems unnecessary.

However, once we have a public [parse method](https://github.com/CharlieEriksen/esprima-dotnet/blob/2f1cc74a748579638d31c344cbee80228a36e8ff/src/Esprima/Ast/BinaryExpression.cs#L51) for binary operators, it would be nice to have a method which converts them back to string. This might be a nice addition.

CharlieEriksen:
Oh doh. It seems like my current version of esprisma was out of date, and didn't have your change here which made that ctor public:
https://github.com/sebastienros/esprima-dotnet/commit/7014c7244c63cc8b953d1a2272840dd15229bf0c#diff-eac8c60fbc85e1f0fe43dfca964b354016f8f90116f2457e84ef41e3aa84528c

Closing, since it is indeed irrelevant with it being public.

PR #287, opened 2022-06-28
Move max assignment depth into the parser options

CharlieEriksen:
I ran into the issue of hitting the max assignment depth in the parser. @adams85 recently refactored some of the options to move the `OnNodeAction` into the Parser Options in https://github.com/sebastienros/esprima-dotnet/pull/283. I figured it'd be appropriate to do the same for the max assignment depth.

I also increased the max depth to 200, as the main config javascript file for AWS hits the 100 limit. I couldn't find info on how to add a fixture. But adding in https://us-east-1.console.aws.amazon.com/config/v2/index.js could be done to exercise this change and max limit.


lahma (approved):
Nice addition, thank you!

CharlieEriksen:
Yep, I did goofed on the merge. I fix that now :)

PR #290, opened 2022-06-30
Rewrite ChildNodesEnumerationHelpersGenerator into incremental generator to improve performance

adams85:
The original source generator implementation is kind of naive: it's practically re-run on every key press when editing any source file in the IDE. Luckily, as I've learned, there's a newer approach ([incremental generators](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md)) to mitigate this problem.

### DIFF
--- a/src/Esprima.SourceGenerators/ChildNodesEnumerationHelpersGenerator.cs
+++ b/src/Esprima.SourceGenerators/ChildNodesEnumerationHelpersGenerator.cs
@@ -1,4 +1,5 @@
-﻿using System.Globalization;
+﻿using System.Collections.Immutable;
+using System.Globalization;
 using System.Text;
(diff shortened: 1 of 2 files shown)


lahma (approved):
Cool stuff, I also played a while back with [incremental generators in Jint](https://github.com/sebastienros/jint/pull/1076). Uses Fluid to generate the output from Liquid templates and utilizes Verify to assert proper output.

adams85:
I'm quite new to the source generator business, haven't done it before. It's an awesome concept for sure, opens up a lot of possibilities for improving coding patterns. Thanks for mentioning Fluid, looks like a great project as well. Combined with source generators can be a nice replacement for T4.

PR #291, opened 2022-07-04
Convert to js

jogibear9988:
rebase of the branch #154


jogibear9988:
@adams85 now you can review

jogibear9988:
maybe now I could also be based on AstVisitor, but I don't know atm... I've only rebased what I've programed long time ago

adams85:
I took a quick glance, so just some random thoughts before going into the details:

* > maybe now I could also be based on AstVisitor, but I don't know atm...

   You definitely should. It wouldn't be wise to duplicate the visitation dispatch logic because that would mean another location to keep in mind when changes are made later on.
* The name `ToJavascriptConverter` is pretty weird. I suggest `AstToJavascriptConverter`, `JavascriptWriter` or something like this.
* Declaring a class for a single extension method (`ToJavascriptConverterExtension`) doesn't look good either. We already have a `NodeExtensions` class, I think that would be a perfect place for the `ToJavascript` method.
* I would keep the public API as minimal as possible, so I'd definitely keep the visitor internal and would introduce some options type. Take a look at `AstJson.Options` for ideas.
* It might also be nice to separate the writing logic from the visitor as it is done in the case of `JsonWriter` and `AstToJsonConverter`. But, anyway, we should provide some API which allows consumers to write into a `TextWriter` because that is the standard way when it comes to generating text in .NET. A `StringBuilder` can always be wrapped in a `TextWriter` (see [`StringWriter`](https://docs.microsoft.com/en-us/dotnet/api/system.io.stringwriter.-ctor?view=net-6.0)) but not the other way around.

Ok, I think this is enough for starters. 😄

adams85:
> I would keep the public API as minimal as possible, so I'd definitely keep the visitor internal and would introduce some options type. Take a look at AstJson.Options for ideas.

On second thoughts, we can't really keep the visitor internal as extensions (like JSX) should be supported too. So I think our best bet is to copy the pattern of `AstJson` and `AstToJsonConverter` (+ `JsxAstJsonConverter`) like
* `AstJson` -> `JavascriptWriter`
* `AstJson.Options` -> `JavascriptWriter.Options`
* `AstToJsonConverter` -> `AstToJavascriptConverter`
etc.

If we did this, `JavascriptWriter` could be a better place for the `ToJavascript` convenience method.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+
+        protected virtual void VisitAwaitExpression(AwaitExpression awaitExpression)
+        {
+            Append("await ");
+            Visit(awaitExpression.Argument);
+        }
+
+        protected virtual void VisitConditionalExpression(ConditionalExpression conditionalExpression)
+        {
+            if (conditionalExpression.Test is AssignmentExpression)
```
adams85, line 1282: What about e.g. `var v = (a, b, c) ? d : e`? Looks like this problem is more widespread than it seems at first glance...

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+                }
+                Visit(property.Value);
+                if (property.Value is not ObjectPattern && ExpressionNeedsBrackets(property.Value))
+                {
+                    Append(")");
+                }
+            }
+        }
+
+        protected virtual void VisitPropertyDefinition(PropertyDefinition propertyDefinition)
```
adams85, line 1243: `propertyDefinition.Decorators` are not handled.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+            {
+                Append(";");
+            }
+            if (forOfStatement.Body is not BlockStatement)
+            {
+                DecreaseIndent();
+            }
+        }
+
+        protected virtual void VisitClassDeclaration(ClassDeclaration classDeclaration)
```
adams85, line 1064: `classDeclaration.Decorators` are not handled.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+                Visit(function.Id);
+            }
+            Append("(");
+            VisitNodeList(function.Params, appendSeperatorString: ",");
+            Append(")");
+            AppendBeautificationSpace();
+            Visit(function.Body);
+        }
+
+        protected virtual void VisitClassExpression(ClassExpression classExpression)
```
adams85, line 841: `classExpression.Decorators` are not handled.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+            }
+        }
+
+        protected virtual void VisitBreakStatement(BreakStatement breakStatement)
+        {
+            if (breakStatement.Label != null)
+            {
+                Visit(breakStatement.Label);
+            }
+            Append("break");
```
adams85, line 1421: Wrong order.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return _sb.ToString();
+        }
+
+        public bool IsAsync(Node node)
```
adams85, line 1482: This method looks unnecessary.

src/Esprima/Utils/ToJavascriptConverter.cs
```diff
@@ -0,0 +1,1580 @@
...
+        {
+            Visit(exportSpecifier.Local);
+            if (exportSpecifier.Local != exportSpecifier.Exported)
+            {
+                Append(" as ");
+                Visit(exportSpecifier.Exported);
+            }
+        }
+
+        protected virtual void VisitImport(Import import)
```
adams85, line 921: `import.Attributes` are not handled.

adams85 (commented):
The feature certainly needs some more polish but there are no major problems apart from one thing: cases where expressions need to be wrapped in parentheses. This seem pretty unreliable yet, so needs some more thought.

lahma:
Thanks @adams85 for going through this with such a detail! Just that we are on same page how to proceed as I know @jogibear9988 might have his hands full with esprima-next too. Some alternatives at least:

1. merge as-is, @adams85 's comments are a TODO list for *someone* willing to do cleanup - we will have technical debt in main
2. @jogibear9988 has the time to improve and does a round with the code improving things that he sees fit, then merge
3. *someone* improves this PR further like above, then merge
4. mystery solution

It's a great feature so I'd hope we can agree on some plan to make sure we get it into main, whether "perfect" or incrementally improving.

adams85:
Since I've already got into this and I happen to have some free time now, I could as well make an attempt on finishing the feature. If it's okay with you, I'd just need write permissions to the PR.

> For me one critical point is, the visitor should be public, and the methods virtual, cause i wanted to overwrite the logic wich some ast nodes are written as javascript

You mean you want to sublass it in *another* assembly? If so, maybe we can handle that using an `InternalsVisibleTo` attribute.

lahma:
> Since I've already got into this and I happen to have some free time now, I could as well make an attempt on finishing the feature. If it's okay with you, I'd just need write permissions to the PR.

As I'm just "a contributor with some merge capabilities", I'm afraid I can't give the permissions. But if @jogibear9988 is OK with it, maybe you can branch from this PR into another PR adding fixes on top. I have no idea how the attribution of changes will flow after that 😬

jogibear9988:
> > Since I've already got into this and I happen to have some free time now, I could as well make an attempt on finishing the feature. If it's okay with you, I'd just need write permissions to the PR.
> 
> As I'm just "a contributor with some merge capabilities", I'm afraid I can't give the permissions. But if @jogibear9988 is OK with it, maybe you can branch from this PR into another PR adding fixes on top. I have no idea how the attribution of changes will flow after that 😬

For me this would be full okay. You also can close, create a new one... as you like. But I also added him to my fork

adams85:
You may try this: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

jogibear9988:
this is checked
image

adams85:
```
remote: Permission to jogibear9988/esprima-dotnet.git denied to adams85.
fatal: unable to access 'https://github.com/jogibear9988/esprima-dotnet/': The requested URL returned error: 403
```

Maybe it's because I'm not added as a maintainer here?

jogibear9988:
@adams85 I've added you to my fork, but you need to accept the invite

jogibear9988:
I could not add you here, I'm also not a maintainer here.

adams85:
Ok, it works now. Thanks!

adams85:
> Don't like this approach. Why should esprima-net have an internals-visible-to my assembly? Why not make it public? Think the best way for people to modify the JS output of some parts would be to override the visitor for the node they wanted to change.

Generally, it's a good idea to minimize the API surface area so we can make modification more freely without breaking consumer code. However, you have a good point here, so the visitor will remain public then.

BTW, what about [this issue](https://github.com/sebastienros/esprima-dotnet/pull/291#discussion_r914207032)? Can `Identifier.Name` ever be null? I suspect it can't. Do any of you know it for sure?

jogibear9988:
if you look here: https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API

name of identifier is string and not nullable

adams85:
Thanks for the clarification. I'll fix this too.

CharlieEriksen:
Wow @jogibear9988, what a beautiful piece of work. I've found myself having to use js-beautify/prettifier in my processing pipeline. Both introduce significant performance issues and are not super reliable. 

This will be epic. I'll try to take a spin on your branch over the weekend and see if it works as-is on my workloads. Fingers crossed!

adams85:
@CharlieEriksen It's a nice work indeed but please not that this is still a WIP. There are known cases where the AST is transformed incorrectly (e.g. operator precedence handling is pretty unreliable currently). The public API is going to change too. Hopefully, in a few days we'll have a maturer version.

adams85:
This was a harder nut to crack than it looked at first sight (or even at second) but finally I proudly present the improved AST to JS conversion! 🎉 (Truth be told, there was so many changes, fixes and improvements that I virtually ended up with a complete rewrite...)

@jogibear9988 Please re-review it, at least the public API, which is mainly defined in the `AstToJavascript` class now. I also override `Node.ToString()` and added a meaningful debugger display format as it's done in the Roslyn AST. Some further things worth mentioning:
* The converter logic was split into two components: `AstToJavascriptConverter` and `JavascriptTextWriter`. The former defines the structure of the output by traversing the AST, the latter is responsible for code formatting. To make this possible, `AstToJavascriptConverter` provides rich context and formatting hints for `JavascriptTextWriter`. (See `JavascriptTextWriter.WriteContext` and the Flags enums).
* We have two formatters currently: the base `JavascriptTextWriter`, which produces minimal output, and the `KnRJavascriptTextWriter` subclass, which implements K&R-style formatting (also provides some knobs to fine tune the formatting rules - see the related `Options` class).

I also adjusted your existing test cases and added another set of tests: I reused the complete test suite 
used for testing the parser. The fixture files get parsed, converted back to text, then re-parsed, finally the two resulting ASTs are compared (only structure and node types for now). So we can say that it's pretty comprehensively tested, however getting parentheses rules right is hard af so I'm still not 100% sure that every corner case and quirk is covered. But I think it should be ok for an initial version.

@CharlieEriksen Now you may give it a go, the unformatted

adams85 (approved):

src/Esprima/Utils/JavascriptTextWriter.cs
```diff
@@ -0,0 +1,375 @@
...
+            case TokenType.RegularExpression:
+                WriteSpace();
+                break;
+            case TokenType.EOF:
+            case TokenType.Punctuator:
+            case TokenType.StringLiteral:
+            case TokenType.Template:
+                break;
+            default:
+                throw new InvalidOperationException();
```
lahma, line 92: should prefer `EsprimaExceptionHelper` for consistency, this is argument out of range I guess?

src/Esprima/Utils/JavascriptTextWriter.cs
```diff
@@ -0,0 +1,375 @@
...
```
lahma, line 138: should prefer `EsprimaExceptionHelper` for consistency, argument out of range?

src/Esprima/Utils/JsonTextWriter.cs
```diff
@@ -53,8 +53,13 @@ private enum StructureKind : byte { Array, Object }
         public JsonTextWriter(TextWriter writer, string? indent)
         {
             _writer = writer ?? ThrowArgumentNullException<TextWriter>(nameof(writer));
-            _writer = writer ?? ThrowArgumentNullException<TextWriter>(nameof(writer));
+
+            if (!string.IsNullOrWhiteSpace(indent))
+            {
+                throw new ArgumentException("Indent must be null or white-space.", nameof(indent));
```
lahma, line 59: should prefer `EsprimaExceptionHelper` for consistency

src/Esprima/Utils/KnRJavascriptTextWriter.cs
```diff
@@ -0,0 +1,480 @@
...
+using Esprima.Ast;
+
+namespace Esprima.Utils;
+
+/// <summary>
+/// Javascript text writer (code formatter) which implements the most common <see href="https://en.wikipedia.org/wiki/Indentation_style#K&amp;R_style">K&amp;R style</see>.
+/// </summary>
+public class KnRJavascriptTextWriter : JavascriptTextWriter
+{
+    public new record class Options : JavascriptTextWriter.Options
```
lahma, line 11: the `new` here feels bad, should this just be `KnRJavascriptTextWriterOptions`? In deal world could just pass suitable `Options` instance to `JavascriptTextWriter` that configures KnR style, but seems that it wouldn't be that easy to implement.

src/Esprima/Utils/KnRJavascriptTextWriter.cs
```diff
@@ -0,0 +1,480 @@
...
+    private int _indentionLevel;
+    private int _currentStatementBodyFlags;
+
+    public KnRJavascriptTextWriter(TextWriter writer, JavascriptTextWriter.Options options) : base(writer, options)
+    {
+        var extendedOptions = options as Options ?? Options.GetDefaultFrom(options);
+
+        if (!string.IsNullOrWhiteSpace(extendedOptions.Indent))
+        {
+            throw new ArgumentException("Indent must be null or white-space.", nameof(extendedOptions));
```
lahma, line 43: `EsprimaExceptionHelper`

src/Esprima/Utils/KnRJavascriptTextWriter.cs
```diff
@@ -0,0 +1,480 @@
...
+        {
+            WriteSpace();
+        }
+        else
+        {
+            base.StartPunctuator(value, flags, context);
+        }
+    }
+
+    #region Arrays
```
lahma, line 186: regions...

lahma (commented):
Wow, you really went the extra mile with this implementation, seems like powerful stuff. I left some comments, mostly about code style as I don't have much other to give in the JS writing area.

src/Esprima/Utils/JavascriptTextWriter.cs
```diff
@@ -0,0 +1,375 @@
...
+        public static readonly Options Default = new();
+    }
+
+    public delegate JavascriptTextWriter Factory(TextWriter writer, Options options);
+
+    private readonly TextWriter _writer;
+
+    public JavascriptTextWriter(TextWriter writer, Options options)
+    {
+        _writer = writer ?? throw new ArgumentNullException(nameof(writer));
```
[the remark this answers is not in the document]
adams85: FYI, `EsprimaExceptionHelper` isn't used consistently throughout the code base.

As a matter of fact, I fail to see why it exists at all (when throw expressions are a thing since C# 7.0). What do we gain from e.g. `Description = description ?? ThrowArgumentNullException<string>(nameof(description));` over `Description = description ?? throw new ArgumentNullException(nameof(description));`? I mean it's not even shorter and you need to type out the generic parameter too. Plus it breaks static analysis. (Not even `DoesNotReturn` solves this completely as the compiler still complains when it's used in a switch case branch.)

So how about getting rid of `EsprimaExceptionHelper` instead? I'm willing to do that if you sign off on it.

and I suggest getting rid of it altogether.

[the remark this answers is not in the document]
lahma: It's for JIT mostly, JIT cannot inline methods that have `throw` in them (call stack would be wrong).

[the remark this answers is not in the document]
adams85: Wow, I learned something today again! Thanks for this valuable info. I'll revise `throw` usage and change it to the helper method where inlining justifies it.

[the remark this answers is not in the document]
lahma: Yeah, method must of course be quite small to be inline candidate (that's why some usages would be more for consistency). Also I *think* that the return of `T` for some throw helpers might not work when wanting the "throw helper pattern", they might need to be void, but not an issue for this PR, just a mental note.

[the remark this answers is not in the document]
adams85: I annotated all my methods with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` which I think it's beneficial to inline. Don't really care whether JIT inlines the rest or not, so in those cases I just use plain `throw`s. As I see, this pattern is followed in other parts of the code (e.g. in `JavascriptParser`). So, are we good or do you insist on using the throw helpers everywhere?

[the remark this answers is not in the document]
lahma: No need, re-evaluate if needed later.

src/Esprima/Utils/JavascriptTextWriter.cs
adams85: It can't really be an argument out of range because we don't switch on an argument here but a class field.  Maybe `NotImplementedException` is something to consider.

src/Esprima/Utils/JavascriptTextWriter.cs
adams85: See https://github.com/sebastienros/esprima-dotnet/pull/291#discussion_r924382111

CharlieEriksen:
Awesome work folks! 

Just for background on my testing, here's how my pipeline works in part:

1. I have some test harnesses that go and browser some large JS application, like the AWS console
2. For each JS file loaded by the browser, that's sent to my code
3. It will parse the JS with esprima-dotnet, to ensure that the "baseline" is parseable, or detect if there's some code that makes it trip up. 
4. It then passes it onto the prettifier, which now first uses this code for prettifying, or fall back to js-beautify or prettier
5. It will then take the prettified code, and parse the code again like in step 3 to make sure the prettifying didn't break anything

I've run it through a part of my normal test set so far. I ran into one file that failed step 5:
[https://a.b.cdn.console.awsstatic.com/e195e08e418def022e7ca90f52a774ddd0b54868d7c7d4162697e865948d6c10/vendor-react.js](https://a.b.cdn.console.awsstatic.com/e195e08e418def022e7ca90f52a774ddd0b54868d7c7d4162697e865948d6c10/vendor-react.js)

It seems to generate the following:
`for (var e = vi, t = e.length, n = "value" in gi ? gi.value : gi.textContent, r = n.length, o = 0; o < t && e[o] === n[o]; o++)`

`prettier` generates:
```
 for (
        var e = vi,
          t = e.length,
          n = ("value" in gi) ? gi.value : gi.textContent,
          r = n.length,
          o = 0;
        o < t && e[o] === n[o];
        o++
      );
```

It seems like it's not putting the conditional expression in parentheses, which means that the `in` statement is invalid.

CharlieEriksen:
I'm also noticing that with the default `beauty` mode enabled, you end up with code that initially get indented in a pretty nice way. But at some point, it seems to give up on indentation for the most part. 
image

adams85:
> It seems to generate the following:
for (var e = vi, t = e.length, n = "value" in gi ? gi.value : gi.textContent, r = n.length, o = 0; o < t && e[o] === n[o]; o++)

This is what I meant by corner cases and quirks... `in` operators needs special care in `for...in` loops because of the syntax ambiguity. I handled this but your tests have showed that plain `for` loops also have this problem. Which is, of course, understandable as the parser doesn't look ahead that much to know if it deals with a plain `for` loop or a `for...in` one... Anyway, I corrected this.

> I'm also noticing that with the default beauty mode enabled, you end up with code that initially get indented in a pretty nice way. But at some point, it seems to give up on indentation for the most part.

Looks like a bug too. Could you send me the problematic code as text?

adams85:
@CharlieEriksen Could you redo your tests? It should work now.

CharlieEriksen:
Tested again, ran flawlessly. And doing a few spot tests, I haven't found any more cases of the formatting being weird. :)

lahma (approved):
Changes look good to me, and we can always fine-tune later if necessary.

lahma:
I've published the latest main on NuGet (including this PR) as version [3.0.0-beta-3](https://www.nuget.org/packages/Esprima/3.0.0-beta-3).

PR #297, opened 2022-07-21
Code style/formatting improvements

adams85:


lahma:
I will be on mobile from now on until Monday,  so harder to check big changes. How does git blame behave after this? I guess every line will be changed?

lahma:
Ignore revs file can help here, but probably after squash merge when hash is known.


https://akrabat.com/ignoring-revisions-with-git-blame/

adams85:
Yeah, it looks like an ignore revs file is the way to go if we want this. Here's another article: https://medium.com/codex/how-to-introduce-a-code-formatter-without-messing-up-git-history-4a16bd074c10

lahma (approved):
Probably invalidates all current PRs, but a good cleanup.

PR #309, opened 2022-08-14
Remove invalid hash addition to private identifier name

lahma:
In other places the private identifier doesn't include the original `#` as name prefix, but `ParseIdentifierOrPrivateIdentifierName()` adds it for some odd reason. The `PrivateIdentifier` container already tells us that this is... a private identifier.

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -1400,7 +1400,9 @@ public partial class JavaScriptParser
             return ThrowUnexpectedToken<Identifier>(token);
         }
 
-        return isPrivateField ? Finalize(node, new PrivateIdentifier('#' + (string?) token.Value)) : Finalize(node, new Identifier((string?) token.Value!));
+        return isPrivateField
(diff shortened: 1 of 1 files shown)


jogibear9988:
?? I don't understand. You added it in the last pull req: https://github.com/sebastienros/esprima-dotnet/pull/236/files#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1R1403

lahma:
I just moved it in order to not need the token field assignment: https://github.com/sebastienros/esprima-dotnet/pull/236/files#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1L1410 , that's when I noticed that it didn't look right.

jogibear9988:
so it seems I introduced this :-)
https://github.com/sebastienros/esprima-dotnet/pull/205/files#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1R1438

jogibear9988:
I will also fix this in esprima next

lahma:
Cool,  thanks checking!

PR #310, opened 2022-08-15
Refinements around tokens and comments

adams85:
Besides removing JSX-awareness from core parser, this PR also proposes moving `ParsedToken`/`ParsedComment` into the `Esprima.Ast` namespace and changing them into node-like types because the information contained by these types belongs to the AST (see e.g. the tree displayed in AST Explorer). Roslyn, for example, models tokens and comments similarly.

### DIFF
--- a/src/Esprima/Ast/Literal.cs
+++ b/src/Esprima/Ast/Literal.cs
@@ -49,18 +49,10 @@ public sealed class Literal : Expression
     public RegexValue? Regex { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
 
     public string? StringValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.StringLiteral ? (string) Value! : null; }
-    public bool BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral && ReferenceEquals(Value, s_boxedTrue); }
-    public double NumericValue
-    {
-        get => TokenType switch
-        {
-            TokenType.NumericLiteral => (double) Value!,
-            TokenType.BooleanLiteral => ReferenceEquals(Value, s_boxedTrue) ? 1 : 0,
-            _ => default
-        };
-    }
+    public bool? BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral ? ReferenceEquals(Value, s_boxedTrue) : null; }
+    public double? NumericValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.NumericLiteral ? (double) Value! : null; }
     public Regex? RegexValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.RegularExpression ? (Regex?) Value : null; }
-    public BigInteger BigIntValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BigIntLiteral ? (BigInteger) Value! : default; }
+    public BigInteger? BigIntValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BigIntLiteral ? (BigInteger) Value! : null; }
 
     internal override Node? NextChildNode(ref ChildNodes.Enumerator enumerator) => null;
 
--- a/src/Esprima/Ast/Literal.cs
+++ b/src/Esprima/Ast/Literal.cs
@@ -7,33 +7,26 @@ namespace Esprima.Ast;
 
 public sealed class Literal : Expression
 {
-    private static readonly object _boxedTrue = true;
-    private static readonly object _boxedFalse = false;
+    private static readonly object s_boxedTrue = true;
+    private static readonly object s_boxedFalse = false;
 
     internal Literal(TokenType tokenType, object? value, string raw) : base(Nodes.Literal)
     {
         TokenType = tokenType;
         Value = value;
         Raw = raw;
-
-        if (tokenType == TokenType.NumericLiteral)
-        {
-            NumericValue = (double) value!;
-        }
     }
 
-    public Literal(string? value, string raw) : this(TokenType.StringLiteral, value, raw)
+    public Literal(string value, string raw) : this(TokenType.StringLiteral, value, raw)
     {
     }
 
-    public Literal(bool value, string raw) : this(TokenType.BooleanLiteral, value ? _boxedTrue : _boxedFalse, raw)
+    public Literal(bool value, string raw) : this(TokenType.BooleanLiteral, value ? s_boxedTrue : s_boxedFalse, raw)
     {
-        NumericValue = value ? 1 : 0;
     }
 
     public Literal(double value, string raw) : this(TokenType.NumericLiteral, value, raw)
     {
-        NumericValue = value;
     }
 
     public Literal(BigInteger value, string raw) : this(TokenType.BigIntLiteral, value, raw)
@@ -55,11 +48,19 @@ public sealed class Literal : Expression
     public string Raw { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
     public RegexValue? Regex { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
 
-    public string? StringValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.StringLiteral ? Value as string : null; }
-    public bool BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral && ReferenceEquals(Value, _boxedTrue); }
-    public double NumericValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public string? StringValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.StringLiteral ? (string) Value! : null; }
+    public bool BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral && ReferenceEquals(Value, s_boxedTrue); }
+    public double NumericValue
+    {
+        get => TokenType switch
+        {
+            TokenType.NumericLiteral => (double) Value!,
+            TokenType.BooleanLiteral => ReferenceEquals(Value, s_boxedTrue) ? 1 : 0,
+            _ => default
+        };
+    }
     public Regex? RegexValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.RegularExpression ? (Regex?) Value : null; }
-    public BigInteger? BigIntValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BigIntLiteral ? (BigInteger?) Value : null; }
+    public BigInteger BigIntValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BigIntLiteral ? (BigInteger) Value! : default; }
 
     internal override Node? NextChildNode(ref ChildNodes.Enumerator enumerator) => null;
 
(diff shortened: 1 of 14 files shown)


lahma:

| Old |ParseProgram|angular-1.2.5|10.948 ms|0.0251 ms|421.8750|203.1250|7,004 KB|
| **New** |	|	| **10.454 ms (-5%)** | **0.0233 ms** | **406.2500 (-4%)** | **156.2500 (-23%)** | **6,892 KB (-2%)** |
| Old |ParseProgram|backbone-1.1.0|1.406 ms|0.0098 ms|64.4531|31.2500|1,055 KB|
| **New** |	|	| **1.376 ms (-2%)** | **0.0040 ms** | **62.5000 (-3%)** | **25.3906 (-19%)** | **1,036 KB (-2%)** |
| Old |ParseProgram|jquery-1.9.1|8.381 ms|0.0200 ms|359.3750|171.8750|6,086 KB|
| **New** |	|	| **8.706 ms (+4%)** | **0.0343 ms** | **359.3750 (0%)** | **171.8750 (0%)** | **5,976 KB (-2%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|13.532 ms|0.0368 ms|546.8750|265.6250|9,108 KB|
| **New** |	|	| **13.337 ms (-1%)** | **0.0272 ms** | **531.2500 (-3%)** | **265.6250 (0%)** | **8,901 KB (-2%)** |
| Old |ParseProgram|mootools-1.4.5|6.851 ms|0.0134 ms|296.8750|101.5625|4,901 KB|
| **New** |	|	| **6.632 ms (-3%)** | **0.0151 ms** | **289.0625 (-3%)** | **109.3750 (+8%)** | **4,812 KB (-2%)** |
| Old |ParseProgram|underscore-1.5.2|1.199 ms|0.0023 ms|58.5938|23.4375|959 KB|
| **New** |	|	| **1.200 ms (0%)** | **0.0020 ms** | **56.6406 (-3%)** | **19.5313 (-17%)** | **943 KB (-2%)** |
| Old |ParseProgram|yui-3.12.0|6.071 ms|0.0111 ms|265.6250|117.1875|4,403 KB|
| **New** |	|	| **6.335 ms (+4%)** | **0.0148 ms** | **257.8125 (-3%)** | **125.0000 (+7%)** | **4,301 KB (-2%)** |

adams85:
Added a further commit (https://github.com/sebastienros/esprima-dotnet/pull/310/commits/8d8b0b398cc5a43831c5b214825672ce32d10d50) for the case where `ParserOptions.Token` is enabled. This keeps the memory footprint of tokens the same as before (by utilizing the otherwise unused internal additional data slot).

@lahma So, do you agree with the proposed changes, especially with https://github.com/sebastienros/esprima-dotnet/pull/310/commits/d09a6c76ee82651c1dc596a9fa87b9c16241a5e0?

src/Esprima/Ast/Literal.cs
```diff
@@ -7,33 +7,26 @@ namespace Esprima.Ast;
 
 public sealed class Literal : Expression
 {
-    private static readonly object _boxedTrue = true;
-    private static readonly object _boxedFalse = false;
+    private static readonly object s_boxedTrue = true;
+    private static readonly object s_boxedFalse = false;
```
lahma, line 11: Now that I remember, I'm not sure if `s_` prefix should be used (editor config might be wrong), Sebastien's other projects don't use it as a suffix

src/Esprima/Ast/Literal.cs
```diff
@@ -55,11 +48,19 @@ public Literal(string pattern, string flags, object? value, string raw) : this(T
     public string Raw { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
     public RegexValue? Regex { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
 
-    public string? StringValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.StringLiteral ? Value as string : null; }
-    public bool BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral && ReferenceEquals(Value, _boxedTrue); }
-    public double NumericValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public string? StringValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.StringLiteral ? (string) Value! : null; }
+    public bool BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral && ReferenceEquals(Value, s_boxedTrue); }
```
lahma, line 52: note own my own doings, the TokenType check for BooleanValue is quite redundant

src/Esprima/Ast/SyntaxComment.cs
```diff
@@ -0,0 +1,40 @@
...
+using Esprima.Utils;
+
+namespace Esprima.Ast;
+
+public sealed class SyntaxComment : SyntaxElement
+{
+    public SyntaxComment(CommentType type, string value)
+    {
+        Type = type;
+
```
lahma, line 11: nit: extra line break

src/Esprima/Ast/SyntaxElement.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    /// Sets additional, user-defined data associated with the specified key.
+    /// </summary>
+    /// <remarks>
+    /// The operation is not guaranteed to be thread-safe. In case concurrent access or update is possible, the necessary synchronization is caller's responsibility.
+    /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void SetAdditionalData(object key, object? value) => _additionalDataContainer.SetData(key, value);
+
+    public Range Range;
```
lahma, line 30: these might be candidates for `ref readonly` returns

src/Esprima/Ast/SyntaxToken.cs
```diff
@@ -0,0 +1,27 @@
...
+        Type = type;
+
+        Value = value;
+        RegexValue = regexValue;
+    }
+
+    public TokenType Type { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+
+    public string Value { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+    public RegexValue? RegexValue
```
lahma, line 18: nit: maybe extra line break in order as RegexValue has more behavior

src/Esprima/JavascriptParser.cs
```diff
@@ -208,10 +208,13 @@ private protected void CollectComments()
             {
                 var e = comments[i];
 
-                e.Value = _scanner.Source.Slice(e.Slice.Start, e.Slice.End);
-                e.Location = new Location(e.Location.Start, e.Location.End, _errorHandler.Source);
+                var comment = new SyntaxComment(e.Type, value: _scanner.Source.Slice(e.Slice.Start, e.Slice.End))
+                {
+                    Range = new Range(e.Start, e.End),
+                    Location = new Location(in e.StartPosition, in e.EndPosition, _errorHandler.Source)
```
lahma, line 214: nit: `in` modifier on call site probably redundant

src/Esprima/JavascriptParser.cs
```diff
@@ -2007,7 +2013,7 @@ private Expression ParseBinaryExpression()
         var allowAndOr = true;
         var allowNullishCoalescing = true;
 
-        void UpdateNullishCoalescingRestrictions(in Token t)
+        static void UpdateNullishCoalescingRestrictions(in Token t, ref bool allowAndOr, ref bool allowNullishCoalescing)
```
lahma, line 2016: nice!

src/Esprima/JsxToken.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Runtime.CompilerServices;
+
+namespace Esprima;
+
+public enum JsxTokenType
+{
+    Unknown,
+    Identifier,
+    Text
+};
```
lahma, line 10: nit: semicolon is redundant

src/Esprima/Loc.cs
```diff
@@ -5,11 +5,11 @@ namespace Esprima;
 [DebuggerDisplay("{ToString()}")]
 public readonly struct Location : IEquatable<Location>
 {
-    public Position Start { get; }
-    public Position End { get; }
-    public string? Source { get; }
+    public readonly Position Start;
```
lahma, line 8: again might be ref-readonly candidates, probably could be benchmarked separately like the previous if makes any difference

lahma (approved):
Looks good, left couple notes, mostly nit-picking.

lahma:
> Added a further commit ([8d8b0b3](https://github.com/sebastienros/esprima-dotnet/commit/8d8b0b398cc5a43831c5b214825672ce32d10d50)) for the case where `ParserOptions.Token` is enabled. This keeps the memory footprint of tokens the same as before (by utilizing the otherwise unused internal additional data slot).
> 
> @lahma So, do you agree with the proposed changes, especially with [d09a6c7](https://github.com/sebastienros/esprima-dotnet/commit/d09a6c76ee82651c1dc596a9fa87b9c16241a5e0)?

I think it looks good, we can always re-iterate when needed 👍🏻

src/Esprima/Ast/Literal.cs
adams85: I changed this because the IDE suggested it. *.editorconfig* has this currently:
```
dotnet_naming_style.static_field_style.required_prefix = s_
```

lahma: Yes, the original was taking from.. was it Roslyn repo and might have had some cruft left behind. I think other repos just don't follow the `s_`, but not a concern for this PR.

src/Esprima/Ast/Literal.cs
adams85: As a matter of fact, `public bool? BooleanValue { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => TokenType == TokenType.BooleanLiteral ? ReferenceEquals(Value, _boxedTrue) : null; }` might be a bit cleaner. It would be a BC but this way it'd be clear whether the literal has a boolean value or not. I'm leaning towards this change. What do you think?

src/Esprima/Ast/SyntaxElement.cs
adams85: We've already [discussed this](https://github.com/sebastienros/esprima-dotnet/pull/273#issue-1268699911). AFAICS, we can't really do any better without massive breaking changes.

src/Esprima/JavascriptParser.cs
adams85: True but I like to be explicit about this.

src/Esprima/Loc.cs
adams85: I changed this to make it consistent with `Range`. Ref readonly with aggressive inlining would probably be just as performant, however, I think public fields just do fine in this case so why complicate it.

src/Esprima/Ast/Literal.cs
lahma: Well from Jint development perspective, we already check the token type to know which property to read value from, so there we know type = BooleanLiteral before accessing the getter so everything else but plain `true/false` is just noise. But I understand the idea behind missing boolean value as this is not a boolean. Currently there's no fast way to get just the reference equals as the boxed value is hidden. But this probably is in the category of micro-optimizations. I'm fine with either way. Jint would just do `BooleanValue.Value` blindly as type is known.

src/Esprima/Ast/SyntaxElement.cs
lahma: Well you have a good memory, unlike I. Location is usually only used on error conditions so doesn't matter that much from Jint's perspective.

src/Esprima/Ast/Literal.cs
adams85: Ok, then I'd say let's do this: https://github.com/sebastienros/esprima-dotnet/pull/310/commits/5184ee1a9f07246408b592d4b92b1726779d2e9d

> Jint would just do BooleanValue.Value blindly as type is known

When you know the type beforehand, you don't need the `XXXValue` properties at all, you can just do the cast directly. E.g. `(bool) Value` in the case of a boolean.

Otherwise, you can read `BooleanValue` and from the result it will be clear whether it actually has a boolean value (`true`/`false`) or not (`null`).

PR #313, opened 2022-08-15
Reduce memory allocations in JavaScriptParser

lahma:
With this PR we are pretty much getting all the allocations from AST + strings. I do have a StringSegment/TextSpan branch somewhere though...

* allow clearing of and supplying initial data directly to ArrayList
* make ParsedParameters a struct
* pool data structures used by ParseBinaryExpression, ParseLexicalBinding and ParsePattern

Also Jint tests pass.

### DIFF
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -89,6 +89,45 @@ internal struct ArrayList<T> : IReadOnlyList<T>
 #endif
     }
 
+    /// <summary>
+    /// Note, expects ownership of the array!
+    /// </summary>
+    public ArrayList(T[] initialData)
+    {
+        if (initialData is null)
+        {
+            ThrowArgumentNullException(nameof(initialData));
+        }
+
+        _items = initialData;
+        _count = initialData.Length;
+
+#if DEBUG
+        _localVersion = 0;
+        _sharedVersion = _count > 0 ? new[] { _localVersion } : null;
+#endif
+    }
+
+    public ArrayList(NodeList<Node> initialData)
+    {
+        if (initialData._items is null)
+        {
+            _items = null;
+            _count = 0;
+            return;
+        }
+
+        _items = new T[initialData.Count];
+        _count = initialData.Count;
+
+        Array.Copy(initialData._items!, 0, _items, 0, _count);
+
+#if DEBUG
+        _localVersion = 0;
+        _sharedVersion = _count > 0 ? new[] { _localVersion } : null;
+#endif
+    }
+
     private int Capacity => _items?.Length ?? 0;
 
     public int Count
@@ -140,6 +179,19 @@ internal struct ArrayList<T> : IReadOnlyList<T>
         Debug.Assert(_items != null);
         _items[_count] = item;
         _count++;
+
+        OnChanged();
+    }
+
+    internal void Clear()
+    {
+        AssertUnchanged();
+
+        if (this._items is not null)
+        {
+            Array.Clear(this._items, 0, this._count);
+            this._count = 0;
+        }
 
         OnChanged();
     }
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -71,7 +71,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
 #if DEBUG
     private int[] _sharedVersion;
     private int _localVersion;
-#endif
+ #endif
 
     public ArrayList(int initialCapacity)
     {
@@ -108,24 +108,25 @@ internal struct ArrayList<T> : IReadOnlyList<T>
 #endif
     }
 
-    public ArrayList(NodeList<Node> initialData)
+    public static ArrayList<Node> FromNodeList(NodeList<Node> initialData)
     {
-        if (initialData._items is null)
+        if (initialData._count == 0)
         {
-            _items = null;
-            _count = 0;
-            return;
+            return new ArrayList<Node>();
         }
 
-        _items = new T[initialData.Count];
-        _count = initialData.Count;
-
-        Array.Copy(initialData._items!, 0, _items, 0, _count);
-
+        var items = new Node[initialData.Count];
+        Array.Copy(initialData._items!, 0, items, 0, initialData._count);
+        return new ArrayList<Node>
+        {
+            _items = items,
+            _count = initialData.Count,
 #if DEBUG
-        _localVersion = 0;
-        _sharedVersion = _count > 0 ? new[] { _localVersion } : null;
+            _localVersion = 0,
+            _sharedVersion = initialData._count > 0 ? new[] { 0 } : null
 #endif
+        };
+
     }
 
     private int Capacity => _items?.Length ?? 0;
@@ -187,13 +188,16 @@ internal struct ArrayList<T> : IReadOnlyList<T>
     {
         AssertUnchanged();
 
-        if (this._items is not null)
+        if (this._count > 0)
         {
             Array.Clear(this._items, 0, this._count);
             this._count = 0;
         }
 
-        OnChanged();
+#if DEBUG
+        _sharedVersion = new [] { 0 };
+        _localVersion = 0;
+#endif
     }
 
     internal void Resize(int size)
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -71,7 +71,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
 #if DEBUG
     private int[] _sharedVersion;
     private int _localVersion;
- #endif
+#endif
 
     public ArrayList(int initialCapacity)
     {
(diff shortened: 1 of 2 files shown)


lahma:

| Old |ParseProgram|angular-1.2.5|10.554 ms|0.0166 ms|421.8750|156.2500|6,892 KB|
| **New** |	|	| **10.617 ms (+1%)** | **0.0244 ms** | **359.3750 (-15%)** | **156.2500 (0%)** | **6,010 KB (-13%)** |
| Old |ParseProgram|backbone-1.1.0|1.388 ms|0.0021 ms|62.5000|25.3906|1,036 KB|
| **New** |	|	| **1.388 ms (0%)** | **0.0101 ms** | **54.6875 (-13%)** | **27.3438 (+8%)** | **913 KB (-12%)** |
| Old |ParseProgram|jquery-1.9.1|8.611 ms|0.0190 ms|359.3750|171.8750|5,976 KB|
| **New** |	|	| **8.396 ms (-2%)** | **0.0208 ms** | **312.5000 (-13%)** | **156.2500 (-9%)** | **5,144 KB (-14%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|13.704 ms|0.0360 ms|531.2500|265.6250|8,901 KB|
| **New** |	|	| **13.033 ms (-5%)** | **0.0201 ms** | **468.7500 (-12%)** | **218.7500 (-18%)** | **7,824 KB (-12%)** |
| Old |ParseProgram|mootools-1.4.5|6.542 ms|0.0114 ms|289.0625|109.3750|4,812 KB|
| **New** |	|	| **6.629 ms (+1%)** | **0.0131 ms** | **250.0000 (-14%)** | **117.1875 (+7%)** | **4,182 KB (-13%)** |
| Old |ParseProgram|underscore-1.5.2|1.174 ms|0.0018 ms|56.6406|19.5313|943 KB|
| **New** |	|	| **1.176 ms (0%)** | **0.0048 ms** | **46.8750 (-17%)** | **19.5313 (0%)** | **792 KB (-16%)** |
| Old |ParseProgram|yui-3.12.0|6.177 ms|0.0132 ms|257.8125|125.0000|4,301 KB|
| **New** |	|	| **6.062 ms (-2%)** | **0.0100 ms** | **234.3750 (-9%)** | **109.3750 (-13%)** | **3,850 KB (-10%)** |

sebastienros:
Was wondering why you didn't share tables anymore ;)

lahma:
This is the current top allocation percentages when parsing `jquery.mobile-1.4.2.js` 100 times with this PR code. 

image

src/Esprima/ArrayList.cs
```diff
@@ -144,6 +183,19 @@ internal void Add(T item)
...
+    {
+        AssertUnchanged();
+
+        if (this._items is not null)
+        {
+            Array.Clear(this._items, 0, this._count);
+            this._count = 0;
+        }
+
+        OnChanged();
```
adams85, line 196: In debug builds this call results in exception when `_sharedVersion` is `null`.

src/Esprima/JavascriptParser.cs
```diff
@@ -3877,17 +3877,20 @@ private RestElement ParseRestElement(ref ArrayList<Token> parameters)
         return Finalize(node, new RestElement(arg));
     }
 
+    // pooled for ParseLexicalBinding calls
+    private ArrayList<Token> _parseFormalParameterParameters;
+
     private void ParseFormalParameter(ref ParsedParameters options)
     {
-        var parameters = new ArrayList<Token>();
+        _parseFormalParameterParameters.Clear();
```
adams85, line 3879: Consider the following input: `function f(x, y = function(a, b) {}, z) {}`

In such cases `ParseFormalParameter` is re-entered. So, without saving the current `_parseFormalParameterParameters` into a local variable and restoring it at the end of the method the pooling would be buggy.

The same goes for `ParseVariableDeclaration` and `ParseLexicalBinding`. (BTW, `ParseCatchClause` and `ParseGroupExpression` are candidates for pooling too).

A further remark: do we really need multiple `ArrayList<Token>` fields (`_parseLexicalBindingParameters`, `_parseFormalParameterParameters`, etc.) for this kind of pooling? A single one looks sufficient me (after ironing out the aforementioned problem, of course).

adams85 (changes requested):

adams85:
Nice improvements again, you're squeezing those allocations like hell! 😄 

BTW, [this benchmark](https://github.com/sebastienros/esprima-dotnet/pull/313#issuecomment-1215379293) of yours showing that `string`s and `Identifier`s are on the top of the list gave me an idea how we might push this even further:

Let's suppose we have the following input:
```js
let abc;
abc = 0;
```

We declare something, then later we reference that using its name. Pretty common stuff. When we parse this, we end up with a bunch of identical `Identifier`s (well, not completely identical ones because of the metadata like location, range, etc.) However, the `Name` property of these identifiers will be exactly the same strings. I mean they'll contain the same characters. They won't be the same *reference* (apart from some special cases covered by `ParserExtensions.Slice`) currently though. But they could be and this is where I see a room for improvement.

It looks possible to introduce some kind of a string pooling by means of which we could make sure that `Scanner` won't allocate the same string multiple times. We may borrow some ideas from [here](https://docs.microsoft.com/en-us/windows/communitytoolkit/high-performance/stringpool) as it looks to do exactly what I'm talking about.

lahma:
Thanks for the review, I'll try to address these things as time permits. Workdays with family can be a bit... crowded...

> BTW, [this benchmark](https://github.com/sebastienros/esprima-dotnet/pull/313#issuecomment-1215379293) of yours showing that `string` and `Identifier` are sitting on the top of the list gave me an idea how we might push this even further:
> 
> Let's suppose we have the following input:
> 
> ```js
> let abc;
> abc = 0;
> ```
> 
> We declare something, then later we reference that using its name. Pretty common stuff. When we parse this, we end up with an AST containing two identical `Identifier`s (well, not completely identical because of metadata like location, range, etc.) However, the `Name` property of these identifiers will be exactly the same strings. I mean they'll contain the same characters. They won't currently be the same _instance_ (apart from some special cases covered by `ParserExtensions.Slice`) though. But they could be and this is where I see a room for improvement.
> 
> It looks possible to introduce some kind of a string pooling by means of which we could make sure that `Scanner` won't allocate the same string multiple times. We may borrow some ideas from [here](https://docs.microsoft.com/en-us/windows/communitytoolkit/high-performance/stringpool) as it looks to do exactly what I'm talking about.

Yes some time ago I was wondering about `Identifier` pooling, which basically dies down when location and range are introduced, like you've said, Jint needs them anyways always for proper problem reporting. I think pooling of string identifiers is a competing solution to `StringSegment/TextSpan` - this requires benchmarking and pros and cons.

I discussed this with Sebastien and ideally we wouldn't introduce a new library for this. Copying existing (MIT or similar) code is OK with attribution as probably headers on top of file.

lahma:
Did a review round fixes, current stats:

| Old |ParseProgram|angular-1.2.5|10.635 ms|0.0835 ms|421.8750|156.2500|6,892 KB|
| **New** |	|	| **10.467 ms (-2%)** | **0.0665 ms** | **343.7500 (-19%)** | **156.2500 (0%)** | **5,751 KB (-17%)** |
| Old |ParseProgram|backbone-1.1.0|1.416 ms|0.0042 ms|62.5000|25.3906|1,036 KB|
| **New** |	|	| **1.381 ms (-2%)** | **0.0113 ms** | **52.7344 (-16%)** | **21.4844 (-15%)** | **888 KB (-14%)** |
| Old |ParseProgram|jquery-1.9.1|8.708 ms|0.0183 ms|359.3750|171.8750|5,976 KB|
| **New** |	|	| **8.263 ms (-5%)** | **0.0391 ms** | **296.8750 (-17%)** | **140.6250 (-18%)** | **4,995 KB (-16%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|13.556 ms|0.0296 ms|531.2500|265.6250|8,901 KB|
| **New** |	|	| **13.379 ms (-1%)** | **0.0428 ms** | **468.7500 (-12%)** | **234.3750 (-12%)** | **7,678 KB (-14%)** |
| Old |ParseProgram|mootools-1.4.5|6.716 ms|0.0149 ms|289.0625|109.3750|4,812 KB|
| **New** |	|	| **6.479 ms (-4%)** | **0.0172 ms** | **242.1875 (-16%)** | **85.9375 (-21%)** | **4,033 KB (-16%)** |
| Old |ParseProgram|underscore-1.5.2|1.207 ms|0.0038 ms|56.6406|19.5313|943 KB|
| **New** |	|	| **1.150 ms (-5%)** | **0.0018 ms** | **44.9219 (-21%)** | **17.5781 (-10%)** | **748 KB (-21%)** |
| Old |ParseProgram|yui-3.12.0|6.143 ms|0.0115 ms|257.8125|125.0000|4,301 KB|
| **New** |	|	| **5.986 ms (-3%)** | **0.0313 ms** | **226.5625 (-12%)** | **85.9375 (-31%)** | **3,792 KB (-12%)** |

adams85:
> Thanks for the review, I'll try to address these things as time permits. Workdays with family can be a bit... crowded...

No problem. Hope you don't mind, I did some further corrections (see https://github.com/sebastienros/esprima-dotnet/pull/313/commits/309a481890c3fb6fff4782ffcc03456477c70246). It was faster for me this way than doing another review iteration. Check out these changes and if you agree, we can go ahead with merging.

> Yes some time ago I was wondering about Identifier pooling, which basically dies down when location and range are introduced, like you've said, Jint needs them anyways always for proper problem reporting. I think pooling of string identifiers is a competing solution to StringSegment/TextSpan - this requires benchmarking and pros and cons.

I discussed this with Sebastien and ideally we wouldn't introduce a new library for this. Copying existing (MIT or similar) code is OK with attribution as probably headers on top of file.

Fully agreed. I'll explore the string pooling approach and will get back to you in case the results look promising.

adams85 (approved):

lahma:
Thanks for the help! A couple kilobyte increase in some benchmarks after latest commit but really insignificant.

PR #314, opened 2022-08-17
Implement string pooling in scanner/parser

adams85:
Motivation:

> BTW, [this benchmark](https://github.com/sebastienros/esprima-dotnet/pull/313#issuecomment-1215379293) of yours showing that `string` and `Identifier` are sitting on the top of the list gave me an idea how we might push this even further:
> 
> Let's suppose we have the following input:
> 
> ```js
> let abc;
> abc = 0;
> ```
> 
> We declare something, then later we reference that using its name. Pretty common stuff. When we parse this, we end up with an AST containing two identical `Identifier`s (well, not completely identical because of metadata like location, range, etc.) However, the `Name` property of these identifiers will be exactly the same strings. I mean they'll contain the same characters. They won't currently be the same _instance_ (apart from some special cases covered by `ParserExtensions.Slice`) though. But they could be and this is where I see a room for improvement.
> 
> It looks possible to introduce some kind of a string pooling by means of which we could make sure that `Scanner` won't allocate the same string multiple times. We may borrow some ideas from [here](https://docs.microsoft.com/en-us/windows/communitytoolkit/high-performance/stringpool) as it looks to do exactly what I'm talking about.

First results:

| Old |ParseProgram|angular-1.2.5|21.162 ms|0.4196 ms|1000.0000|468.7500|-|5,744 KB|
| **New** |	|	| **20.491 ms (-3%)** | **1.2497 ms** | **937.5000 (-6%)** | **437.5000 (-7%)** | **-** | **5,553 KB (-3%)** |
| Old |ParseProgram|backbone-1.1.0|2.609 ms|0.4331 ms|160.1563|70.3125|-|888 KB|
| **New** |	|	| **2.706 ms (+4%)** | **0.0774 ms** | **148.4375 (-7%)** | **66.4063 (-6%)** | **-** | **867 KB (-2%)** |
| Old |ParseProgram|jquery-1.9.1|15.677 ms|1.6023 ms|812.5000|375.0000|-|4,991 KB|
| **New** |	|	| **16.503 ms (+5%)** | **2.2575 ms** | **812.5000 (0%)** | **375.0000 (0%)** | **-** | **4,854 KB (-3%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|27.572 ms|29.7640 ms|1250.0000|625.0000|93.7500|7,677 KB|
| **New** |	|	| **25.710 ms (-7%)** | **5.6449 ms** | **1218.7500 (-3%)** | **593.7500 (-5%)** | **31.2500 (-67%)** | **7,349 KB (-4%)** |
| Old |ParseProgram|mootools-1.4.5|12.870 ms|1.5982 ms|703.1250|312.5000|-|4,028 KB|
| **New** |	|	| **13.191 ms (+2%)** | **0.9527 ms** | **671.8750 (-4%)** | **312.5000 (0%)** | **-** | **3,936 KB (-2%)** |
| Old |ParseProgram|underscore-1.5.2|2.197 ms|0.1964 ms|140.6250|62.5000|-|748 KB|
| **New** |	|	| **2.251 ms (+2%)** | **0.1881 ms** | **140.6250 (0%)** | **62.5000 (0%)** | **-** | **739 KB (-1%)** |
| Old |ParseProgram|yui-3.12.0|11.652 ms|0.8847 ms|656.2500|281.2500|-|3,789 KB|
| **New** |	|	| **11.899 ms (+2%)** | **0.6330 ms** | **609.3750 (-7%)** | **296.8750 (+6%)** | **-** | **3,620 KB (-4%)** |

These results are not as good as I expected. But there are some parameters to tweak (string pool initial buffer size, buffer growth rate, etc) and the hash algorithm can be optimized further as well. We'll see if it can get any better.

### DIFF
--- a/src/Esprima/ParserExtensions.cs
+++ b/src/Esprima/ParserExtensions.cs
@@ -26,6 +26,13 @@ public static class ParserExtensions
             Scanner.TryGetInternedPunctuator(source);
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static string Slice(this string source, int start, int end, ref StringPool stringPool)
+    {
+        var sourceSpan = source.AsSpan(start, end - start);
+        return TryGetInternedString(sourceSpan) ?? stringPool.GetOrCreate(sourceSpan);
+    }
+
     public static string Slice(this string source, int start, int end)
     {
         var len = source.Length;
--- a/src/Esprima/ParserExtensions.cs
+++ b/src/Esprima/ParserExtensions.cs
@@ -2,7 +2,7 @@
 
 namespace Esprima;
 
-public static class ParserExtensions
+internal static class ParserExtensions
 {
     private static readonly string[] s_charToString = new string[256];
 
@@ -33,24 +33,8 @@ public static class ParserExtensions
         return TryGetInternedString(sourceSpan) ?? stringPool.GetOrCreate(sourceSpan);
     }
 
-    public static string Slice(this string source, int start, int end)
-    {
-        var len = source.Length;
-        var from = start < 0 ? Math.Max(len + start, 0) : Math.Min(start, len);
-        var to = end < 0 ? Math.Max(len + end, 0) : Math.Min(end, len);
-        var span = Math.Max(to - from, 0);
-
-        if (span == 1)
-        {
-            return CharToString(source[from]);
-        }
-
-        var substring = TryGetInternedString(source.AsSpan(from, span)) ?? source.Substring(from, span);
-        return substring;
-    }
-
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public static string CharToString(char c)
+    internal static string CharToString(char c)
     {
         if (c >= 0 && c < s_charToString.Length)
         {
@@ -61,7 +45,7 @@ public static class ParserExtensions
     }
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public static char CharCodeAt(this string source, int index)
+    internal static char CharCodeAt(this string source, int index)
     {
         if (index < 0 || index > source.Length - 1)
         {
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -208,7 +208,7 @@ public partial class JavaScriptParser
             {
                 var e = comments[i];
 
-                var comment = new SyntaxComment(e.Type, value: _scanner.Source.Slice(e.Slice.Start, e.Slice.End))
+                var comment = new SyntaxComment(e.Type, value: _scanner.Source.Slice(e.Slice.Start, e.Slice.End, ref _scanner._stringPool))
                 {
                     Range = new Range(e.Start, e.End),
                     Location = new Location(in e.StartPosition, in e.EndPosition, _errorHandler.Source)
@@ -224,7 +224,7 @@ public partial class JavaScriptParser
     /// </summary>
     private protected string GetTokenRaw(in Token token)
     {
-        return _scanner.Source.Slice(token.Start, token.End);
+        return _scanner.Source.Slice(token.Start, token.End, ref _scanner._stringPool);
     }
 
     private protected SyntaxToken ConvertToken(in Token token)
@@ -4165,7 +4165,7 @@ public partial class JavaScriptParser
         var expr = ParseExpression();
         if (expr.Type == Nodes.Literal)
         {
-            directive = _scanner.Source.Slice(token.Start + 1, token.End - 1);
+            directive = _scanner.Source.Slice(token.Start + 1, token.End - 1, ref _scanner._stringPool);
         }
 
         ConsumeSemicolon();
--- a/src/Esprima/JsxParser.cs
+++ b/src/Esprima/JsxParser.cs
@@ -482,7 +482,7 @@ public class JsxParser : JavaScriptParser
                 }
             }
 
-            var id = _scanner.Source.Slice(start, _scanner.Index);
+            var id = _scanner.Source.Slice(start, _scanner.Index, ref _scanner._stringPool);
             return JsxToken.CreateIdentifier(id, start, end: _scanner.Index, _scanner.LineNumber, _scanner.LineStart);
         }
 
--- a/src/Esprima/Scanner.cs
+++ b/src/Esprima/Scanner.cs
@@ -48,6 +48,8 @@ public sealed partial class Scanner
     private List<string> _curlyStack;
     private readonly StringBuilder strb = new();
 
+    internal StringPool _stringPool;
+
     private static int HexValue(char ch)
     {
         if (ch >= 'A')
@@ -594,7 +596,7 @@ public sealed partial class Scanner
             }
         }
 
-        return Source.Slice(start, Index);
+        return Source.Slice(start, Index, ref _stringPool);
     }
 
     private string GetComplexIdentifier()
@@ -1574,7 +1576,7 @@ public sealed partial class Scanner
             _curlyStack.RemoveAt(_curlyStack.Count - 1);
         }
 
-        var rawTemplate = Source.Slice(start + 1, Index - rawOffset);
+        var rawTemplate = Source.Slice(start + 1, Index - rawOffset, ref _stringPool);
         var value = notEscapeSequenceHead == default ? cooked.ToString() : null;
 
         return Token.CreateTemplate(cooked: value, rawTemplate, head, tail, notEscapeSequenceHead, start, end: Index, LineNumber, LineStart);
@@ -2428,6 +2430,7 @@ public sealed partial class Scanner
             _curlyStack.Capacity = 0;
             strb.Clear();
             strb.Capacity = 0;
+            _stringPool = default;
 
             return Token.CreateEof(Index, LineNumber, LineStart);
         }
--- a/src/Esprima/StringPool.cs
+++ b/src/Esprima/StringPool.cs
@@ -3,6 +3,9 @@ using System.Runtime.CompilerServices;
 
 namespace Esprima;
 
+/// <summary>
+/// A heavily slimmed down version of <see cref="HashSet{String}"/> which can be used to reduce memory allocations when dissecting a string.
+/// </summary>
 // Based on:
 // * https://github.com/dotnet/runtime/blob/v6.0.8/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
 // * https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/v7.1.2/Microsoft.Toolkit.HighPerformance/Buffers/StringPool.cs
@@ -11,6 +14,8 @@ internal struct StringPool
     private int[]? _buckets;
     private Entry[]? _entries;
     private int _count;
+
+    public int Count { [MethodImpl(MethodImplOptions.AggressiveInlining)] get => _count; }
 
     /// <summary>
     /// Initializes buckets and slots arrays. Uses suggested capacity by finding next prime
--- a/src/Esprima/StringPool.cs
+++ b/src/Esprima/StringPool.cs
@@ -1,0 +1,196 @@
+﻿using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Esprima;
+
+// Based on:
+// * https://github.com/dotnet/runtime/blob/v6.0.8/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
+// * https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/v7.1.2/Microsoft.Toolkit.HighPerformance/Buffers/StringPool.cs
+internal struct StringPool
+{
+    private int[]? _buckets;
+    private Entry[]? _entries;
+    private int _count;
+
+    /// <summary>
+    /// Initializes buckets and slots arrays. Uses suggested capacity by finding next prime
+    /// greater than or equal to capacity.
+    /// </summary>
+    private int Initialize(int capacity)
+    {
+        int size = capacity;
+        var buckets = new int[size];
+        var entries = new Entry[size];
+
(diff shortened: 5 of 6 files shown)


src/Esprima/ParserExtensions.cs
```diff
@@ -15,617 +16,631 @@ static ParserExtensions()
...
 
-                if (source[from + 1] == 'r' && source[from + 2] == 'g')
+                if (source[1] == 'r' && source[2] == 'g')
                 {
                     return "arg";
                 }
             }
 
-            if (source[from] == 'f' && source[from + 1] == 'o' && source[from + 2] == 'r')
+            if (source[0] == 'f' && source[1] == 'o' && source[2] == 'r')
```
sebastienros, line 89: Commenting on this PR though I understand it's unrelated, was just wondering what techniques were used here. 

`Span.IndexOf()` is vectorized now, so checking each char should be slower, and is harder to read.

src/Esprima/ParserExtensions.cs
```diff
@@ -15,617 +16,631 @@ static ParserExtensions()
...
 
-        var substring = source.Substring(from, span);
+        return null;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static string Slice(this string source, int start, int end, ref StringPool stringPool)
+    {
+        var sourceSpan = source.AsSpan(start, end - start);
+        return TryGetInternedString(sourceSpan) ?? stringPool.GetOrCreate(sourceSpan);
```
sebastienros, line 633: Couldn't the keywords in TryGetInternedString be part of the stringPool too? That might make things simpler, and probably as fast.

sebastienros:
The TextSpan strategy might have a disadvantage compared to interned strings in the way that the source of the script will always be referenced by the final AST once the parsing is done. In terms of memory there is no allocations, but it might hide the fact that a bigger chunk of memory can't be freed. Micro-benchmarks won't show that. Might be interesting to have a column with memory delta before/after the method and a GC to show that.

One could argue that it's using more memory (after parsing), but it's still faster and still allocating less.

sebastienros:
FYI, Ben Adams made an interning library https://github.com/benaadams/Ben.StringIntern

src/Esprima/ParserExtensions.cs
adams85: https://github.com/sebastienros/esprima-dotnet/pull/317 may improve the situation.

src/Esprima/ParserExtensions.cs
adams85: Good point. In the case of string pooling it doesn't make sense indeed to hard-code a bunch of keywords. What seems to make sense is to use interned string instances of keywords that are hard-coded anyway (i.e. the keywords defined in the scanner). It would be redundant to duplicate those by placing them into the string pool.

adams85:
> FYI, Ben Adams made an interning library https://github.com/benaadams/Ben.StringIntern

Thanks for the tip. I've checked it out and found that the implementation is based on BCL's `HashSet<T>` just like ours but it's a more generic one than that and does *much* more than what we need here. Which means much higher complexity, so I doubt we'd get better results from switching to this lib.

One thing we may consider borrowing from there is [the Marvin alg](https://github.com/benaadams/Ben.StringIntern/blob/716cfbba5338ce070aafbeafbdd549395649f575/src/Helpers/Marvin.cs#L12) for hashing `ReadOnlySpan<char>` (but this would involve pulling in `System.Runtime.CompilerServices.Unsafe`). However, [this alg](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/winui-7.1.2/CommunityToolkit.HighPerformance/Helpers/Internals/SpanHelper.Hash.cs#L87) looks even more powerful. This will definitely need some benchmarking.

> The TextSpan strategy might have a disadvantage compared to interned strings in the way that **the source of the script will always be referenced by the final AST once the parsing is done**.

The fact that the entire input would be kept around (which could be take up megabytes) until even a single node which has a reference to it is alive looks like a showstopper to me.

I see two major problems with the TextSpan approach:
1. Our parser doesn't collect whitespace trivia like e.g. Roslyn, so in the case of larger files we'd keep around a lot of unreferenced whitespace unnecessary. (And this goes for tokens and comments when collection of them is turned off in the options.)
2. Even if our parser collected whitespace trivia, the concept would begin to fall apart as soon as we consider AST rewriting/modification. When a large part of the tree is rewritten, we can easily end up with the unfortunate situation outlined above.

@lahma What do you think about these arguments? Am I missing here something maybe?

adams85:
I've found some low hanging fruits around `Scanner.ScanPunctuator`:

| Old |ParseProgram|angular-1.2.5|21.056 ms|2.2627 ms|1000.0000|468.7500|-|5,744 KB|
| **New** |	|	| **19.723 ms (-6%)** | **1.2730 ms** | **718.7500 (-28%)** | **343.7500 (-27%)** | **-** | **4,391 KB (-24%)** |
| Old |ParseProgram|backbone-1.1.0|2.595 ms|0.1273 ms|164.0625|78.1250|-|888 KB|
| **New** |	|	| **2.759 ms (+6%)** | **0.4853 ms** | **140.6250 (-14%)** | **42.9688 (-45%)** | **-** | **679 KB (-24%)** |
| Old |ParseProgram|jquery-1.9.1|15.487 ms|0.4387 ms|875.0000|437.5000|-|4,991 KB|
| **New** |	|	| **15.798 ms (+2%)** | **0.8879 ms** | **625.0000 (-29%)** | **312.5000 (-29%)** | **-** | **3,770 KB (-24%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|27.276 ms|28.5918 ms|1312.5000|656.2500|187.5000|7,677 KB|
| **New** |	|	| **24.657 ms (-10%)** | **0.5061 ms** | **1000.0000 (-24%)** | **500.0000 (-24%)** | **31.2500 (-83%)** | **5,800 KB (-24%)** |
| Old |ParseProgram|mootools-1.4.5|12.754 ms|1.5094 ms|703.1250|312.5000|-|4,028 KB|
| **New** |	|	| **12.707 ms (0%)** | **4.7469 ms** | **500.0000 (-29%)** | **250.0000 (-20%)** | **-** | **3,095 KB (-23%)** |
| Old |ParseProgram|underscore-1.5.2|2.370 ms|0.0625 ms|132.8125|54.6875|-|748 KB|
| **New** |	|	| **2.290 ms (-3%)** | **0.2174 ms** | **121.0938 (-9%)** | **39.0625 (-29%)** | **-** | **574 KB (-23%)** |
| Old |ParseProgram|yui-3.12.0|11.833 ms|2.5972 ms|656.2500|296.8750|-|3,789 KB|
| **New** |	|	| **11.487 ms (-3%)** | **1.0623 ms** | **468.7500 (-29%)** | **234.3750 (-21%)** | **-** | **2,853 KB (-25%)** |

It looks now we're getting somewhere. There are still some knobs to tweak but I expect that memory allocation can't significantly be reduced further. I'm more optimistic about execution time because of the more advanced ROS hashing methods.

Also, the latest commit includes improvements regarding https://github.com/sebastienros/esprima-dotnet/pull/314#discussion_r948267245

lahma:
> > FYI, Ben Adams made an interning library https://github.com/benaadams/Ben.StringIntern
> 
> Thanks for the tip. I've checked it out and found that the implementation is based on BCL's `HashSet<T>` just like ours but it's a more generic one than that and does _much_ more than what we need here. Which means much higher complexity, so I doubt we'd get better results from switching to this lib.
> 
> One thing we may consider borrowing from there is [the Marvin alg](https://github.com/benaadams/Ben.StringIntern/blob/716cfbba5338ce070aafbeafbdd549395649f575/src/Helpers/Marvin.cs#L12) for hashing `ReadOnlySpan<char>` (but this would involve pulling in `System.Runtime.CompilerServices.Unsafe`). However, [this alg](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/winui-7.1.2/CommunityToolkit.HighPerformance/Helpers/Internals/SpanHelper.Hash.cs#L87) looks even more powerful. This will definitely need some benchmarking.

Those are cool implementations. I actually tried the WindowsCommunityToolkit version briefly on Jint side a while back, can't remember specifics though 😄 

> > The TextSpan strategy might have a disadvantage compared to interned strings in the way that **the source of the script will always be referenced by the final AST once the parsing is done**.
> 
> The fact that the entire input would be kept around (which could be take up megabytes) until even a single node which has a reference to it is alive looks like a showstopper to me.
> 
> I see two major problems with the TextSpan approach:
> 
> 1. Our parser doesn't collect whitespace trivia like e.g. Roslyn, so in the case of larger files we'd keep around a lot of unreferenced whitespace unnecessary. (And this goes for tokens and comments when collection of them is turned off in the options.)
> 2. Even if our parser collected whitespace trivia, the concept would begin to fall apart as soon as we consider AST rewriting/modification. When a large part of the tree is rewritten, we can easily end up with the unfortunate situation outlined above.
> 
> @lahma What do you think about these arguments? Am I missing here something maybe?

This is *a very good* point. I'm starting to lean towards string interning + recently used pooling approach that you have here. Maybe we'll ditch the StringSegment approach unless something really interesting comes along later.

Is ScanPunctuator work something that could be brough to main separately to focus on the differences the pooling in this PR brings, would it make sense? Some of the diffs look familiar like char to string lookup, I probably also experimented with those on my branch.

adams85:
> Is ScanPunctuator work something that could be brough to main separately to focus on the differences the pooling in this PR brings, would it make sense? 

It would make perfect sense. Will do it soon.

adams85:
So, it seems this is what (a plain and simple) string pooling can bring to the table:

| Old |ParseProgram|angular-1.2.5|19.870 ms|0.9480 ms|750.0000|375.0000|-|4,718 KB|
| **New** |	|	| **19.869 ms (0%)** | **1.5634 ms** | **750.0000 (0%)** | **343.7500 (-8%)** | **-** | **4,392 KB (-7%)** |
| Old |ParseProgram|backbone-1.1.0|2.730 ms|0.1545 ms|140.6250|58.5938|-|721 KB|
| **New** |	|	| **2.848 ms (+4%)** | **0.5838 ms** | **140.6250 (0%)** | **46.8750 (-20%)** | **-** | **679 KB (-6%)** |
| Old |ParseProgram|jquery-1.9.1|15.617 ms|3.7405 ms|656.2500|328.1250|-|4,079 KB|
| **New** |	|	| **15.717 ms (+1%)** | **3.2395 ms** | **593.7500 (-10%)** | **281.2500 (-14%)** | **-** | **3,770 KB (-8%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|24.085 ms|0.8815 ms|1031.2500|500.0000|-|6,304 KB|
| **New** |	|	| **24.907 ms (+3%)** | **3.6787 ms** | **1000.0000 (-3%)** | **468.7500 (-6%)** | **31.2500** | **5,800 KB (-8%)** |
| Old |ParseProgram|mootools-1.4.5|12.483 ms|1.4284 ms|546.8750|265.6250|-|3,290 KB|
| **New** |	|	| **12.588 ms (+1%)** | **1.0525 ms** | **500.0000 (-9%)** | **250.0000 (-6%)** | **-** | **3,095 KB (-6%)** |
| Old |ParseProgram|underscore-1.5.2|2.261 ms|0.0589 ms|113.2813|42.9688|-|603 KB|
| **New** |	|	| **2.331 ms (+3%)** | **0.0859 ms** | **117.1875 (+3%)** | **39.0625 (-9%)** | **-** | **574 KB (-5%)** |
| Old |ParseProgram|yui-3.12.0|11.954 ms|2.6612 ms|500.0000|250.0000|-|3,079 KB|
| **New** |	|	| **11.726 ms (-2%)** | **0.0871 ms** | **484.3750 (-3%)** | **234.3750 (-6%)** | **-** | **2,900 KB (-6%)** |

At least, I couldn't improve it further. I tested both Marvin and Djb algorithms for ROS hashing but it didn't make any difference. Probably because the most frequent pooled strings tend to be short.

As to @sebastienros's remark in the other PR:

> Note that parlot falls in this category at the moment. It makes me think that some allocations benchmarks are flawed since we don't take the preallocated memory into account when measuring an iteration.

> I haven't looked at the lifetime scope of the pool either in the intern pr. It matters in concurrent systems where the same pool might be in locking contention.

This implementation doesn't use pre-allocated memory. Each `Scanner` instance [starts with an own empty pool](https://github.com/sebastienros/esprima-dotnet/blob/26f3ce9a5bcc3e9232652b3c1075bea81bf49414/src/Esprima/Scanner.cs#L51) and [gets rid of that as soon as EOF is reached](https://github.com/sebastienros/esprima-dotnet/blob/26f3ce9a5bcc3e9232652b3c1075bea81bf49414/src/Esprima/Scanner.cs#L2433) (so the buffer can be GC'd even if the `Scanner` instance is kept alive.)

Because of this, we have no concurrency worries either. (A shared pool, however, is not something I'd consider because that looks pretty tricky and it's pretty dubious that we would have a clear gain from that in the end.)

Now it's up to you whether to merge this or not.

src/Esprima/ParserExtensions.cs
```diff
@@ -28,6 +28,13 @@ static ParserExtensions()
...
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static string Slice(this string source, int start, int end, ref StringPool stringPool)
+    {
+        var sourceSpan = source.AsSpan(start, end - start);
+        return TryGetInternedString(sourceSpan) ?? stringPool.GetOrCreate(sourceSpan);
+    }
+
     public static string Slice(this string source, int start, int end)
```
lahma, line 38: I think this is no longer used and can probably be removed? seems that also `CharToString` should probably by internal, maybe even `CharCodeAt`?

lahma (approved):
I think there's a nice balance in memory benefits and small (negligible) perf hit. Probably should just take this in and see if we can refine later (hard-coded dictionary etc).

src/Esprima/ParserExtensions.cs
adams85: I agree. I think we can make the whole `ParserExtensions` class internal for now.

PR #317, opened 2022-08-18
Clean up generation logic and output of StringMatcher

adams85:
This PR aims to improve the situation mentioned [here](https://github.com/sebastienros/esprima-dotnet/pull/314#discussion_r948265709) by utilizing the string matcher source generator.

Also does a bit of cleanup around that.

### DIFF
--- a/src/Esprima.SourceGenerators/Helpers/SourceGenerationHelper.cs
+++ b/src/Esprima.SourceGenerators/Helpers/SourceGenerationHelper.cs
@@ -58,16 +58,16 @@ internal class StringMatcherAttribute : System.Attribute
             }
         }
 
+        if (checkNull)
+        {
+            sb.Append(indentStr).AppendLine("    if (input is null)");
+            sb.Append(indentStr).AppendLine("    {");
+            sb.Append(indentStr).Append("        return ").Append(returnString ? "null" : "false").AppendLine(";");
+            sb.Append(indentStr).AppendLine("    }");
+        }
+
         if (byLength.Length > 1)
         {
-            if (checkNull)
-            {
-                sb.Append(indentStr).AppendLine("    if (input is null)");
-                sb.Append(indentStr).AppendLine("    {");
-                sb.Append(indentStr).Append("        return ").Append(returnString ? "null" : "false").AppendLine(";");
-                sb.Append(indentStr).AppendLine("    }");
-            }
-
             sb.Append(indentStr).AppendLine("    switch (input.Length)");
             sb.Append(indentStr).AppendLine("    {");
         }
@@ -79,16 +79,25 @@ internal class StringMatcherAttribute : System.Attribute
                 sb.Append(indentStr).Append("        case ").Append(group.Key).AppendLine(":");
             }
 
-            var discriminatorIndex = group.Count() > 1 ? FindDiscriminatorIndex(group) : -1;
-
-            if (discriminatorIndex != -1)
+            if (group.Count() == 1)
             {
-                sb.Append(indentStr).Append(baseIndent).Append("    var disc").Append(group.Key).Append(" = input[").Append(discriminatorIndex).AppendLine("];");
+                var item = group.First();
+                sb.Append(indentStr).Append(baseIndent).Append("    return ");
+                StringEquality(sb, item);
+                if (returnString)
+                {
+                    sb.Append(" ? ").Append("\"").Append(Escape(item)).Append("\"").Append(" : null");
+                }
+                sb.AppendLine(";");
+                continue;
             }
-            else if (group.Count() > 1)
+
+            var discriminatorIndex = FindDiscriminatorIndex(group);
+
+            if (discriminatorIndex == -1)
             {
                 // hash-based or equality-based then, let compiler generate decision tree
-                var switchIndent = indentStr + indentStr + baseIndent;
+                var switchIndent = byLength.Length > 1 ? indentStr + indentStr + baseIndent : indentStr + baseIndent;
                 sb.Append(switchIndent).AppendLine("switch (input)");
                 sb.Append(switchIndent).AppendLine("{");
                 if (returnString)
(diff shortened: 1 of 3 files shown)


lahma:
Seems that this has the same outcome as my weekend experiment, things get slower. Oddly there's no memory benefit either.

| Old |ParseProgram|angular-1.2.5|10.470 ms|0.0248 ms|343.7500|156.2500|5,744 KB|
| **New** |	|	| **11.052 ms (+6%)** | **0.0511 ms** | **343.7500 (0%)** | **156.2500 (0%)** | **5,744 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|1.393 ms|0.0070 ms|52.7344|21.4844|888 KB|
| **New** |	|	| **1.510 ms (+8%)** | **0.0067 ms** | **52.7344 (0%)** | **21.4844 (0%)** | **888 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|8.244 ms|0.0167 ms|296.8750|140.6250|4,991 KB|
| **New** |	|	| **8.791 ms (+7%)** | **0.0159 ms** | **296.8750 (0%)** | **140.6250 (0%)** | **4,991 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|13.221 ms|0.0358 ms|468.7500|234.3750|7,677 KB|
| **New** |	|	| **14.005 ms (+6%)** | **0.0451 ms** | **468.7500 (0%)** | **234.3750 (0%)** | **7,677 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|6.600 ms|0.0088 ms|242.1875|117.1875|4,028 KB|
| **New** |	|	| **7.130 ms (+8%)** | **0.0224 ms** | **242.1875 (0%)** | **117.1875 (0%)** | **4,028 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|1.173 ms|0.0063 ms|44.9219|19.5313|748 KB|
| **New** |	|	| **1.278 ms (+9%)** | **0.0016 ms** | **44.9219 (0%)** | **19.5313 (0%)** | **748 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|5.974 ms|0.0084 ms|226.5625|85.9375|3,789 KB|
| **New** |	|	| **6.406 ms (+7%)** | **0.0222 ms** | **226.5625 (0%)** | **85.9375 (0%)** | **3,789 KB (0%)** |

lahma:
For reference, this is the dictionary I used, it contains all JS keywords too which were missing, like `async`.

```csharp
    [StringMatcher(
        "if", "in", "do", "var", "for", "new", "try", "let", "this", "else", "case", "void", "with", "enum", "await",
        "while", "break", "catch", "throw", "const", "yield", "class", "super", "return", "typeof", "delete", "switch",
        "export", "import", "default", "finally", "extends", "function", "continue", "debugger", "instanceof", "async", "static",
        "undefined", "true", "false", "null", "get", "set",
        // some common ones in our test data set (benchmarks + test suite)
        "length", "object", "Object", "obj", "Array", "Math", "data", "done", "args", "arguments", "Symbol", "prototype",
        "options", "value", "name", "self", "&&", "||", "===", "!===", "key", "\"use strict\"", "use strict"
    )]
    private static partial string? Slice(ReadOnlySpan<char> span);
```

src/Esprima.SourceGenerators/Helpers/SourceGenerationHelper.cs
```diff
@@ -58,16 +58,16 @@ void StringEquality(StringBuilder builder, string toCheck)
             }
         }
 
-        if (byLength.Length > 1)
```
lahma, line 61: Nice cleanup all around!

lahma:
This is a competing solution for both your earlier pooling and my (unfinished StringSegment/TextSpan). One potential I know that is here, is the fact that we should be able to match keywords etc with *reference equality* as long as they are part of the dictionary. So instead of  `token.Type == Keyword && "let".Equals(token.Value)` we could write `token.Type == Keyword && ReferenceEquals("let", token.Value)` because the string should be interned by the compiler.

This solution will not reduce memory pressure, but *could* improve string equality checks which are pretty much the bread and butter. In my StringSegment PR it's easy to see that perf is getting slower with spanification, but memory usage takes jump downwards.

adams85:
> This is a competing solution for both your earlier pooling and my (unfinished StringSegment/TextSpan).

This wasn't intended as a solution to the string allocation problem but merely as an attempt on improving the `ParserExtensions.Slice` function. According to your benchmark, unfortunately, it fails to do so: although code was simplified, execution time got worse. So I'll drop the first commit and let's merge only the other one which does some cleanup around the string matcher generator logic and output.

> One potential I know that is here, is the fact that we should be able to match keywords etc with reference equality as long as they are part of the dictionary. So instead of token.Type == Keyword && "let".Equals(token.Value) we could write token.Type == Keyword && ReferenceEquals("let", token.Value) because the string should be interned by the compiler.

Wouldn't this rather apply to the string pooling PR? I mean `token.Type == Keyword && "let".Equals(token.Value)` makes sense only when we've made sure that `token.Value` is an interned string instance (in the case of keywords)?

> EDIT: there's also a path that acorn uses, add extra enum flag to token to directly say it's "let", "async" etc at parsing time when say, keyword is checked. This way during parsing it would be just number equality.

Not sure that it would worth the hassle if we used `ReferenceEquals` as you mentioned above.

src/Esprima.SourceGenerators/Helpers/SourceGenerationHelper.cs
adams85: Thanks!

lahma (approved):

lahma:
> This wasn't intended as a solution to the string allocation problem but merely as an attempt on improving the `ParserExtensions.Slice` function. According to your benchmark, unfortunately, it fails to do so: although code was simplified, execution time got worse. So I'll drop the first commit and let's merge only the other one which does some cleanup around the string matcher generator logic and output.

Yes, the code is a low cleaner, a good nerd-snipe for me to investigate which paths actually make slower than all the indexing - my guess would be large switches (above 6 targets) which translate to hash code.

> > One potential I know that is here, is the fact that we should be able to match keywords etc with reference equality as long as they are part of the dictionary. So instead of token.Type == Keyword && "let".Equals(token.Value) we could write token.Type == Keyword && ReferenceEquals("let", token.Value) because the string should be interned by the compiler.
> 
> Wouldn't this rather apply to the string pooling PR? I mean `token.Type == Keyword && "let".Equals(token.Value)` makes sense only when we've made sure that `token.Value` is an interned string instance (in the case of keywords)?

When `Slice` is used and dictionary contains all targets, we should get interned result and that is what is set into tokens I believe.

> > EDIT: there's also a path that acorn uses, add extra enum flag to token to directly say it's "let", "async" etc at parsing time when say, keyword is checked. This way during parsing it would be just number equality.
> 
> Not sure that it would worth the hassle if we used `ReferenceEquals` as you mentioned above.

Yes, again a competing solution. The enum flag would also work in world where we use spans instead of substrings / string instances.

adams85:
> Yes, the code is a low cleaner, a good nerd-snipe for me to investigate which paths actually make slower than all the indexing - my guess would be large switches (above 6 targets) which translate to hash code.

I'm almost sure that large switches on span parameters cause the slowdown. There's some compiler magic which generates hash for these spans. Looks like character-based comparison beats that (at least, in our case).

PR #323, opened 2022-08-20
Implement Parse/TryParse in Range, Position and Location

adams85:
Also does some cleanup in these types.

@lahma Could you check that making ctors internal (and replacing them with `From` factory methods in the public API) is ok from Jint's perspective?

@CharlieEriksen Please you also take a look. Would this be acceptable to you?

### DIFF
--- a/src/Esprima/Loc.cs
+++ b/src/Esprima/Loc.cs
@@ -1,29 +1,52 @@
-﻿using System.Diagnostics;
-using System.Globalization;
+﻿using System.Globalization;
+using System.Runtime.CompilerServices;
 
 namespace Esprima;
 
-[DebuggerDisplay("{ToString()}")]
 public readonly struct Location : IEquatable<Location>
 {
     public readonly Position Start;
     public readonly Position End;
     public readonly string? Source;
 
-    public Location(in Position start, in Position end) : this(start, end, null)
+    private static bool Validate(in Position start, in Position end, bool @throw = true)
+    {
+        if (start == default && end != default)
+        {
+            if (@throw)
+            {
+                EsprimaExceptionHelper.ThrowArgumentOutOfRangeException(nameof(start), start, Exception<ArgumentOutOfRangeException>.DefaultMessage);
+            }
+            return false;
+        }
+
+        if (end == default ? start != default : end < start)
+        {
+            if (@throw)
+            {
+                EsprimaExceptionHelper.ThrowArgumentOutOfRangeException(nameof(end), end, Exception<ArgumentOutOfRangeException>.DefaultMessage);
+            }
+            return false;
+        }
+
+        return true;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Location From(in Position start, in Position end, string? source = null)
+    {
+        Validate(in start, in end);
+        return new Location(start, end, source);
+    }
+
+    internal Location(in Position start, in Position end) : this(start, end, null)
     {
     }
 
-    public Location(in Position start, in Position end, string? source)
+    internal Location(in Position start, in Position end, string? source)
     {
 #if LOCATION_ASSERTS
-        if (start == default && end != default
-            || end == default && start != default
-            || end.Line < start.Line
-            || start.Line > 0 && start.Line == end.Line && end.Column < start.Column)
-        {
-            EsprimaExceptionHelper.ThrowArgumentOutOfRangeException(nameof(end), end, Exception<ArgumentOutOfRangeException>.DefaultMessage);
-        }
+        Validate(in start, in end);
 #endif
 
         Start = start;
@@ -46,6 +69,16 @@ public readonly struct Location : IEquatable<Location>
         return Start.Equals(other.Start)
                && End.Equals(other.End)
                && string.Equals(Source, other.Source);
+    }
+
+    public static bool operator ==(in Location left, in Location right)
+    {
+        return left.Equals(right);
+    }
+
+    public static bool operator !=(in Location left, in Location right)
+    {
+        return !left.Equals(right);
     }
 
     public override int GetHashCode()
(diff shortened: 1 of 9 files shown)


src/Esprima/Loc.cs
```diff
@@ -1,29 +1,52 @@
...
 
-[DebuggerDisplay("{ToString()}")]
 public readonly struct Location : IEquatable<Location>
 {
     public readonly Position Start;
     public readonly Position End;
     public readonly string? Source;
 
-    public Location(in Position start, in Position end) : this(start, end, null)
+    private static bool Validate(in Position start, in Position end, bool @throw = true)
```
lahma, line 12: instead of reserved word, `throwOnError` would probably be better (like in `Type.GetType`)

src/Esprima/Loc.cs
```diff
@@ -74,16 +107,125 @@ public override string ToString()
...
+    private static bool TryParse(ReadOnlySpan<char> s, bool throwIfInvalid, out Location result)
     {
-        return left.Equals(right);
+        int i;
+        if (s[0] != '[' || (i = s.IndexOf(')')) < 0 || ++i < 5)
+        {
+            goto InvalidFormat;
+        }
+
+        var source = s.Slice(i);
```
lahma, line 118: no sure if slicer is best because it's optimized for character strings, this is numeric, but probably not in hot path

src/Esprima/Loc.cs
```diff
@@ -74,16 +107,125 @@ public override string ToString()
...
     }
 
-    public static bool operator !=(in Location left, in Location right)
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool TryParse(ReadOnlySpan<char> s, out Location result) => TryParse(s, throwIfInvalid: false, out result);
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool TryParse(string s, out Location result) => TryParse(s.AsSpan(), out result);
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
lahma, line 220: cannot inline as there's a `throw`

src/Esprima/Position.cs
```diff
@@ -11,24 +12,47 @@ namespace Esprima;
...
-public readonly struct Position : IEquatable<Position>
+public readonly struct Position : IEquatable<Position>, IComparable<Position>
 {
-    public int Line { get; }
-    public int Column { get; }
+    public readonly int Line;
+    public readonly int Column;
 
-    public Position(int line, int column)
+    private static bool Validate(int line, int column, bool @throw = true)
```
lahma, line 20: `throwOnError`

src/Esprima/Position.cs
```diff
@@ -55,16 +119,55 @@ public override string ToString()
                + Column.ToString(CultureInfo.InvariantCulture);
     }
 
-    public static bool operator ==(Position left, Position right)
+    private static bool TryParse(ReadOnlySpan<char> s, bool throwIfInvalid, out Position result)
```
lahma, line 122: maybe could use `throwOnError` as name here too

lahma (commented):
Made required changes (`From` pattern) on Jint side and tests are green. Jint was actually using `CharCodeAt` (easy to copy) and `Slice` (changed to substring) in its `JsonParser`.

CharlieEriksen:
@adams85, this is some banging work. I had low-key been considering adding `ICompareable` to `Position` as well, but you hit it out of the park!

FWIW, I'm personally not a massive fan of `throwOnError` and similar. It adds noise into signatures, and throwing on something like this is a massive anti-pattern in my book. A `Try`-prefixed function is always there to avoid any exception side-effects. So it feels a bit odd to me. But this may be very subjective :) 

The only thing I've also been wanting to add is the ability to do like:

```csharp
var pos = new Position(0, 10);
var loc = new Location(new Position(0, 0), new Position(0, 15));
var loc2 = new Location(new Position(0, 5), new Position(0, 10));
pos.IsInside(loc);
loc2.InInside(loc);
```

But that should now be super easy to do! :D Thanks again!

src/Esprima/Loc.cs
adams85: Hmm, I'm not sure I get this. This `Slice` is not our pooling slice but a plain old ROS `Slice`...

src/Esprima/Loc.cs
adams85: Nice catch!

adams85:
@lahma 
> Made required changes (From pattern) on Jint side and tests are green. Jint was actually using CharCodeAt (easy to copy) and Slice (changed to substring) in its JsonParser.

Please note that, `From` has a minor perf. penalty because of argument validation. If it's on a hot path, you'd probably want to revert to the ctor. Of course, that would require an `InternalsVisibleTo` for Jint.

@CharlieEriksen 
> FWIW, I'm personally not a massive fan of throwOnError and similar. It adds noise into signatures, and throwing on something like this is a massive anti-pattern in my book. A Try-prefixed function is always there to avoid any exception side-effects. So it feels a bit odd to me.

You're absolutely right about this *in general*. However, in this case `throwOnError` used only in private methods out of necessity because other solutions would be even more awkward. I've renamed these overloads to `TryParseCore` though to be more clear about the intentions.

> The only thing I've also been wanting to add is the ability to do like:

Feel free to submit a PR. ;)

> If you can enable sponsorships on your profile or have a Paypal address in your profile, I'd love to be able to send you a small token of my gratitude :)

That'd be very nice of you! Unfortunately, I haven't setup sponsorships yet but [this](https://www.buymeacoffee.com/adams85) can work for you maybe.

src/Esprima/Loc.cs
lahma: Ah sorry, my bad. Span it is.

lahma (approved):

adams85:
@CharlieEriksen Wow, you're being very generous! Thank you for your support.

Issue #300, opened 2022-07-26
A few additional rough edges to polish for v3


adams85:
IMO these are some features/improvements which would be nice to include in v3 (to make it complete, so to speak):

* [x] Support for comments. (The parsing part is already there, comments just need to be collected and exposed to the consumer. There's [ongoing work](https://github.com/sebastienros/esprima-dotnet/pull/178) related this, just need to be finalized AFAICS. Also, ability of writing comments should be added to `JavascriptTextWriter` even if comments were not implemented in `AstToJavascript` for now.)
* [x] Revising and probably changing the design choice of `Node.Data`. It looks more and more like we'd be better off with some dictionary-like field instead of a plain object. The reason behind this is that not only the consumer but also the parser may want to attach additional information to specific nodes (for example, if we happen to implement `attachComments`). Using normal properties for such extra features would unnecessary increase the memory footprint of AST in normal cases. However, we could kill two birds with one stone using a dictionary.
* [x] Writing back JSX to code. (This is more or less done, I'll just need to finalize it.)
* [x] Revising the public API of the parser. (This work is basically [done](https://github.com/sebastienros/esprima-dotnet/pull/175), just need to be rebased and updated.)

Also, I'd appreciate any feedback on these thoughts. If you agree, I can work on these as time allows.

CharlieEriksen:
On the second point, here's what I'm using for data:
```csharp
    public static object? GetData(this Node node, string type)
    {
        if (node.Data is null) return null;
        if (node.Data is not Dictionary<string, object> data) throw new FormatException("The Data field is polluted");

        data.TryGetValue(type, out var value);
        return value;
    }

    public static Node SetData(this Node node, string type, object value)
    {
        node.Data ??= new Dictionary<string, object>();

        if (node.Data is Dictionary<string, object> dictData)
            dictData[type] = value;
        return node;
    }
```

So while I clearly like the option, I'm wondering if there's any risk of it being a slippery slope?

adams85:
The fact you've ended up with such a use of `Node.Data` just confirms that this change would be desirable. I have an API like this in mind:

```csharp
public abstract class Node
{
    // ...

    // removed: public object? Data; 

    public object? GetAdditionalData(object key) { /* ... */ } // could as well be generic to save a cast for consumers
    public void SetAdditionalData(object key, object? value) { /* ... */ }
}
```

[...]

PR #325, opened 2022-08-21
Clean up string matcher output in cases where no single-character discriminator exists

adams85:


test/Esprima.Tests.SourceGenerators/Properties/launchSettings.json
```diff
@@ -0,0 +1 @@
+{}
```
lahma, line 1: seems redundant?

src/Esprima.SourceGenerators/StringMatcherGenerator.cs
```diff
@@ -131,14 +131,34 @@ private static void Execute(SourceProductionContext context, ImmutableArray<Stri
...
+            }
+
             var (typeModifiers, type) = typeGrouping.Key;
             sourceBuilder.Append(typeModifiers).Append(" partial class ").AppendLine(type)
                 .AppendLine("{");
 
+            bool firstMethod = true;
             foreach (var method in typeGrouping)
             {
+                if (firstMethod)
```
lahma, line 153: nit: could just set to false after AppendLine

lahma (approved):
This is what I would call successful nerd-snipe, thanks for the cleanup!

test/Esprima.Tests.SourceGenerators/Properties/launchSettings.json
adams85: Yep, added accidentally.

adams85:
Yeah, I managed to go full nerd with this indeed... 😄 But it felt good, so np. However, I'm running out of free time, so I can't promise pumping energy into the project with the same intensity as recently. I have [one more task](https://github.com/sebastienros/esprima-dotnet/issues/300) which I'd like to do for v3 if possible. How much would this interfere with your work? May I go ahead with this?

lahma:
> Yeah, I managed to go full nerd with this indeed... 😄 But it felt good, so np. However, I'm running out of free time, so I can't promise pumping energy into the project with the same intensity as recently. 

And a big thank you for the latest sprint!

>I have [one more task](https://github.com/sebastienros/esprima-dotnet/issues/300) which I'd like to do for v3 if possible. How much would this interfere with your work? May I go ahead with this?

Fire at will, I'll manage the merge conflicts. You get things done which is more important than my forever-branches 😄

PR #329, opened 2022-08-24
Minor code quality/consistency improvements

adams85:
Some additional cleanup (but no public APIs were harmed in the process 😄)

Also did a few micro-optimizations along the way. I'm not sure about [this one](https://github.com/sebastienros/esprima-dotnet/commit/b0fb4229fedf18edf5766df11b07313e2087ec44). It works fine, I just don't know if you're ok with such techniques or consider it too hacky?


src/Esprima/Ast/NodeList.cs
```diff
@@ -32,6 +42,24 @@ public int Count
         get => _count;
     }
 
+    public T this[int index]
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get
+        {
+            // Following trick can reduce the range check by one
+            if ((uint) index >= (uint) _count)
```
lahma, line 51: nit: I think this is usually written as `(uint) index < (uint) _count` because we want to have hot path as something that matches first

src/Esprima/Ast/AssignmentExpression.cs
```diff
@@ -65,7 +64,7 @@ public static AssignmentOperator ParseAssignmentOperator(string op)
             "??=" => AssignmentOperator.NullishAssign,
             "&&=" => AssignmentOperator.AndAssign,
             "||=" => AssignmentOperator.OrAssign,
-            _ => ThrowArgumentOutOfRangeException<AssignmentOperator>(nameof(op), "Invalid assignment operator: " + op)
+            _ => throw new ArgumentOutOfRangeException(nameof(op), "Invalid assignment operator: " + op)
```
lahma, line 67: for larghe IL such block seems reasonable as they are non-inlineable

src/Esprima/Esprima.csproj
```diff
@@ -43,16 +43,15 @@
 
   <!-- For checking generated output-->
 
-  <!--
   <PropertyGroup>
```
lahma, line 46: are these changes really needed?

src/Esprima/EsprimaExceptionHelper.cs
```diff
@@ -7,17 +7,17 @@ namespace Esprima;
...
-    {
-        throw new IndexOutOfRangeException();
-    }
-
-    [DoesNotReturn]
-    public static T ThrowArgumentOutOfRangeException<T>(string paramName, object actualValue, string? message = null)
-    {
-        throw new ArgumentOutOfRangeException(paramName, actualValue, message);
-    }
+    //[DoesNotReturn]
```
lahma, line 10: commented-out code is bad, delete if not needed

src/Esprima/JavascriptParser.cs
```diff
@@ -80,6 +78,54 @@ public void ReleaseLargeBuffers()
...
+        parseArrayInitializer = BuildCachedDelegateFor(dummyInstance.ParseArrayInitializer);
+        parseObjectInitializer = BuildCachedDelegateFor(dummyInstance.ParseObjectInitializer);
+        parseBinaryExpression = BuildCachedDelegateFor(dummyInstance.ParseBinaryExpression);
+        parseLeftHandSideExpression = BuildCachedDelegateFor(dummyInstance.ParseLeftHandSideExpression);
+        parseLeftHandSideExpressionAllowCall = BuildCachedDelegateFor(dummyInstance.ParseLeftHandSideExpressionAllowCall);
+        parseStatement = BuildCachedDelegateFor(dummyInstance.ParseStatement);
+        parseFunctionSourceElements = BuildCachedDelegateFor(dummyInstance.ParseFunctionSourceElements);
+        parseAsyncArgument = BuildCachedDelegateFor(dummyInstance.ParseAsyncArgument);
+
+        static Func<JavaScriptParser, T> BuildCachedDelegateFor<T>(Func<T> func)
```
lahma, line 120: this is new to me, interesting to see the perf improvements

lahma (commented):
Can you add before/after diffs with BenchmarkDotNet.ResultDiff kinda way for FileParsing so we know nothing went south, otherwise look like nice improvements!

src/Esprima/Ast/NodeList.cs
adams85: I used to make the hot path the first branch until I read or heard somewhere that it doesn't really make a difference with today's CPUs because of branch prediction. Have I got it wrong?

Of course, I'm going to invert it, just being curious.

src/Esprima/Ast/AssignmentExpression.cs
adams85: I think we marked all the methods which can benefit from inlining. In those methods we use helpers, in other methods it doesn't make a difference but I suggest using simple `throw`s as being cleaner.

src/Esprima/Esprima.csproj
adams85: I'd got a bunch of annoying CS0121 warnings in the latest VS 2022 before I made this change. It seemed to fix the issue but I'm not 100% sure. However, it doesn't do any harm if it didn't.

src/Esprima/JavascriptParser.cs
adams85: It's a micro-optimization (parser instances weigh 120 bytes less and initialization of delegates is a one-time cost). I don't expect much improvement from this (especially using the existing benchmarks).

adams85:
> Can you add before/after diffs with BenchmarkDotNet.ResultDiff kinda way for FileParsing so we know nothing went south, otherwise look like nice improvements!

Sure:

| Old |ParseProgram|angular-1.2.5|18.524 ms|0.2234 ms|687.5000|343.7500|-|4,314 KB|
| **New** |	|	| **18.329 ms (-1%)** | **0.0681 ms** | **718.7500 (+5%)** | **343.7500 (0%)** | **-** | **4,314 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|2.473 ms|0.0219 ms|140.6250|27.3438|-|676 KB|
| **New** |	|	| **2.509 ms (+1%)** | **0.0280 ms** | **140.6250 (0%)** | **27.3438 (0%)** | **-** | **676 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|14.417 ms|0.0676 ms|625.0000|312.5000|-|3,765 KB|
| **New** |	|	| **14.596 ms (+1%)** | **0.1079 ms** | **640.6250 (+3%)** | **312.5000 (0%)** | **-** | **3,765 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|23.022 ms|0.3184 ms|1000.0000|468.7500|31.2500|5,800 KB|
| **New** |	|	| **23.129 ms (0%)** | **0.0653 ms** | **1000.0000 (0%)** | **468.7500 (0%)** | **31.2500 (0%)** | **5,800 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|11.746 ms|0.0836 ms|500.0000|250.0000|-|3,090 KB|
| **New** |	|	| **11.812 ms (+1%)** | **0.0537 ms** | **500.0000 (0%)** | **250.0000 (0%)** | **-** | **3,090 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|2.101 ms|0.0262 ms|117.1875|42.9688|-|571 KB|
| **New** |	|	| **2.098 ms (0%)** | **0.0225 ms** | **117.1875 (0%)** | **42.9688 (0%)** | **-** | **571 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|10.815 ms|0.0654 ms|468.7500|234.3750|-|2,857 KB|
| **New** |	|	| **10.810 ms (0%)** | **0.0654 ms** | **468.7500 (0%)** | **234.3750 (0%)** | **-** | **2,857 KB (0%)** |

As expected, no significant changes.

src/Esprima/Ast/NodeList.cs
lahma: Honestly,  I'm not sure.  It's just nyt gut reaction to try to guide to shortest path.

src/Esprima/ArrayList.cs
```diff
@@ -120,7 +150,62 @@ public int Count
...
-    internal void AddRange<TSource>(ArrayList<TSource> list) where TSource : T
+    public T this[int index]
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get
+        {
+            AssertUnchanged();
+
+            // Following trick can reduce the range check by one
+            if ((uint) index < (uint) _count)
```
lahma, line 161: To avoid check I think you first need to create a local array reference, foreach does the same.

src/Esprima/ArrayList.cs
```diff
@@ -120,7 +150,62 @@ public int Count
...
+            return ThrowArgumentOutOfRangeException<T>(nameof(index), index, null);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        set
+        {
+            AssertUnchanged();
+
+            // Following trick can reduce the range check by one
+            if ((uint) index < (uint) _count)
```
lahma, line 175: Same here, local ref to array and check. Length instead of count

lahma (approved):
Couple notes, reading from mobile so might have missed something,  basically LGTM.

src/Esprima/ArrayList.cs
adams85: I'm afraid that nothing's left to improve here. This piece of code was obviously borrowed from [here](https://github.com/dotnet/runtime/blob/v6.0.8/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L143).

We can't really check `items.Length` because it's not necessarily equal to `_count` (may be greater than that). I think the comment about the trick means that using the `uint` cast we need one comparison instead of two (i.e. `index < 0 || index >= _count`).

PR #333, opened 2022-08-29
A few additional minor corrections

adams85:
This PR started out as an enhancement of the sample project to make it (more) useful (#332). But along the way I found some missing parts to implement, issues to fix (and APIs to break... 😄) so I've decided that it's better off split into two parts.

BCs were made especially around error handling/reporting:
* IMO it doesn't make sense to collect exceptions in `CollectingErrorHandler` once we have a separate data class (`ParseError`) for describing errors. This would only result in unnecessary memory allocation since the exception class (`ParserException`) adds no value, just wraps the error class. I set this straight and added a method named `ToException` to convert error objects to throwable exception instances.
* There was a discrepancy between the definition of the type `Position` and its usage in `ParseError.Position`: it's documented that `Position` uses a zero-based column index. The type is used accordingly throughout the codebase except for `ParseError.Position`, which used one-based column indices. I fixed this but kept `ParseError.Column` one-based to not break everything unnecessary. I also documented this for clarity.

I also revised passing of `Position`, `Range` and `Location` parameters. Removed passing `Position` and `Range` structs by-ref as it doesn't really make sense to pass pointer-sized structs that way. On 64-bit platforms it just looks like an unnecessary indirection (unless inlined), plus it may increase the stack consumption of the caller method.

Of course, for the sake of certainty, I checked for regression in performance:

| Old |ParseProgram|angular-1.2.5|17.267 ms|0.0734 ms|750.0000|312.5000|-|4,315 KB|
| **New** |	|	| **17.106 ms (-1%)** | **0.0697 ms** | **750.0000 (0%)** | **312.5000 (0%)** | **-** | **4,315 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|2.322 ms|0.0103 ms|128.9063|35.1563|-|676 KB|
| **New** |	|	| **2.305 ms (-1%)** | **0.0160 ms** | **125.0000 (-3%)** | **46.8750 (+33%)** | **-** | **676 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|13.673 ms|0.0733 ms|640.6250|312.5000|-|3,764 KB|
| **New** |	|	| **13.587 ms (-1%)** | **0.0398 ms** | **625.0000 (-2%)** | **312.5000 (0%)** | **-** | **3,764 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|21.975 ms|0.1064 ms|968.7500|437.5000|-|5,798 KB|
| **New** |	|	| **21.659 ms (-1%)** | **0.2019 ms** | **1031.2500 (+6%)** | **500.0000 (+14%)** | **31.2500** | **5,798 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|10.938 ms|0.0175 ms|500.0000|250.0000|-|3,091 KB|
| **New** |	|	| **10.887 ms (0%)** | **0.0145 ms** | **500.0000 (0%)** | **250.0000 (0%)** | **-** | **3,091 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|1.946 ms|0.0118 ms|113.2813|46.8750|-|571 KB|
| **New** |	|	| **1.930 ms (-1%)** | **0.0093 ms** | **113.2813 (0%)** | **46.8750 (0%)** | **-** | **571 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|9.830 ms|0.0397 ms|468.7500|234.3750|-|2,856 KB|
| **New** |	|	| **9.966 ms (+1%)** | **0.0367 ms** | **468.7500 (0%)** | **234.3750 (0%)** | **-** | **2,856 KB (0%)** |

With these changes we gained a few extra stack frames though: max. nesting depth has been increased from 585 to 589.

### DIFF
--- a/src/Esprima/Ast/Jsx/JsxSyntaxToken.cs
+++ b/src/Esprima/Ast/Jsx/JsxSyntaxToken.cs
@@ -1,0 +1,13 @@
+﻿using System.Runtime.CompilerServices;
+
+namespace Esprima.Ast.Jsx;
+
+public sealed class JsxSyntaxToken : SyntaxToken
+{
+    public JsxSyntaxToken(JsxTokenType type, string value) : base(TokenType.Extension, value)
+    {
+        Type = type;
+    }
+
+    public new JsxTokenType Type { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
+}
--- a/src/Esprima/ParseError.cs
+++ b/src/Esprima/ParseError.cs
@@ -3,20 +3,25 @@
 public sealed class ParseError
 {
     public string Description { get; }
+
     public string? Source { get; }
 
     public bool IsIndexDefined => Index >= 0;
+
     /// <summary>
-    /// Zero-based index within the parsed code string. (Can be negative if code is not available.)
+    /// Zero-based index within the parsed code string. (Can be negative if location information is available.)
     /// </summary>
     public int Index { get; }
 
     public bool IsPositionDefined => Position.Line > 0;
+
     public Position Position { get; }
+
     /// <summary>
-    /// One-based line number. (Can be zero if code is not available.)
+    /// One-based line number. (Can be zero if location information is not available.)
     /// </summary>
     public int LineNumber => Position.Line;
+
     /// <summary>
     /// One-based column index.
     /// </summary>
--- a/src/Esprima/ParseError.cs
+++ b/src/Esprima/ParseError.cs
@@ -22,7 +22,7 @@ public sealed class ParseError
     /// </summary>
     public int Column => Position.Column + 1;
 
-    public ParseError(string description, string? source = null, int index = -1, in Position position = default)
+    public ParseError(string description, string? source = null, int index = -1, Position position = default)
     {
         Description = description ?? throw new ArgumentNullException(nameof(description));
         Source = source;
--- a/src/Esprima/ParseError.cs
+++ b/src/Esprima/ParseError.cs
@@ -6,12 +6,21 @@ public sealed class ParseError
     public string? Source { get; }
 
     public bool IsIndexDefined => Index >= 0;
+    /// <summary>
+    /// Zero-based index within the parsed code string. (Can be negative if code is not available.)
+    /// </summary>
     public int Index { get; }
 
     public bool IsPositionDefined => Position.Line > 0;
     public Position Position { get; }
+    /// <summary>
+    /// One-based line number. (Can be zero if code is not available.)
+    /// </summary>
     public int LineNumber => Position.Line;
-    public int Column => Position.Column;
+    /// <summary>
+    /// One-based column index.
+    /// </summary>
+    public int Column => Position.Column + 1;
 
     public ParseError(string description, string? source = null, int index = -1, in Position position = default)
     {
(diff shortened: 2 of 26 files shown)


adams85:
I was also strongly considering introducing a dedicated options class for `Scanner` because it's pretty weird that its public ctors accept a `ParserOptions` instance, even though that class contains options which are irrelevant to `Scanner`. We could do this without a severe BC if we created a `ScannerOptions` class or something and inherit `ParserOptions` from it. What do you think about this?

src/Esprima/ParseError.cs
```diff
@@ -6,14 +6,23 @@ public sealed class ParseError
     public string? Source { get; }
 
     public bool IsIndexDefined => Index >= 0;
+    /// <summary>
```
lahma, line 9: nit: missing newline before member with documentation

src/Esprima/ParseError.cs
```diff
@@ -6,14 +6,23 @@ public sealed class ParseError
...
+    /// Zero-based index within the parsed code string. (Can be negative if code is not available.)
+    /// </summary>
     public int Index { get; }
 
     public bool IsPositionDefined => Position.Line > 0;
     public Position Position { get; }
+    /// <summary>
```
lahma, line 16: nit: newline

src/Esprima/ParserException.cs
```diff
@@ -5,10 +5,19 @@ public sealed class ParserException : Exception
     public ParseError? Error { get; }
 
     public string? Description => Error?.Description;
-    public string? SourceText => Error?.Source;
+    public string? SourceLocation => Error?.Source;
```
lahma, line 8: nit: newlines to separate (if GH is showing the truth about missing between)

lahma (approved):
Couldn't find more than nit-picking again...

> I was also strongly considering introducing a dedicated options class for Scanner because it's pretty weird that its public ctors accept a ParserOptions instance, even though that class contains options which are irrelevant to Scanner. We could do this without a severe BC if we created a ScannerOptions class or something and inherit ParserOptions from it. What do you think about this?

Might as well to be separate things to have scanner and parser options (without inheritance). They are considered separate things during AST producing anyway.

src/Esprima/Ast/Jsx/JsxSyntaxToken.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Esprima.Ast.Jsx;
+
+public sealed class JsxSyntaxToken : SyntaxToken
+{
+    public JsxSyntaxToken(JsxTokenType type, string value) : base(TokenType.Extension, value)
+    {
+        Type = type;
+    }
+
+    public new JsxTokenType Type { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; }
```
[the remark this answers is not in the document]
adams85: Probably it isn't but `AggressiveInlining` used throughout the AST types, so I'd keep it for consistency. Probably makes no difference but should do no harm either.

adams85:
> Might as well to be separate things to have scanner and parser options (without inheritance). They are considered separate things during AST producing anyway.

Are you ok with [this solution] or you'd do it in another way (like composition instead of internal interface)?

lahma:
I'm good, we can always refine if needed 😉

PR #337, opened 2022-09-07
Add SyntaxElement internal data set and get API

lahma:
In order for Jint to easily leverage the new data slot for its selfish needs, I think it's easiest to open a public API. Exposing internals via attribute doesn't pan out very well as there are clashes with all the attributes that have been internalized in both projects, like `DoesNotReturn` etc.

I think it's enough to add enough documentation and no-no to the description for people to understand that this API isn't for the faint-hearted.

### DIFF
--- a/src/Esprima/Ast/SyntaxElement.cs
+++ b/src/Esprima/Ast/SyntaxElement.cs
@@ -27,6 +27,26 @@ public abstract class SyntaxElement
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public void SetAdditionalData(object key, object? value) => _additionalDataContainer.SetData(key, value);
 
+    /// <summary>
+    /// Returns internal data state value without any checks.
+    /// </summary>
+    /// <remarks>
+    /// This is an advanced API and can/will break additional data handling, you should prefer using SetAdditionalData/SetAdditionalData.
+    /// The operation is not guaranteed to be thread-safe. In case concurrent access or update is possible, the necessary synchronization is caller's responsibility.
+    /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public object? GetInternalData() => _additionalDataContainer._data;
+
+    /// <summary>
+    /// Forcefully sets internal data state value to given value overwriting any current state.
+    /// </summary>
+    /// <remarks>
+    /// This is an advanced API and can/will break additional data handling, you should prefer using SetAdditionalData/SetAdditionalData.
+    /// The operation is not guaranteed to be thread-safe. In case concurrent access or update is possible, the necessary synchronization is caller's responsibility.
+    /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void SetInternalData(object? value) => _additionalDataContainer._data = value;
+
     public Range Range;
     public Location Location;
 
--- a/src/Esprima/Utils/AdditionalDataContainer.cs
+++ b/src/Esprima/Utils/AdditionalDataContainer.cs
@@ -4,7 +4,7 @@ internal struct AdditionalDataContainer
 {
     private static readonly object s_internalDataKey = new();
 
-    private object? _data;
+    internal object? _data;
 
     internal object? InternalData
     {


adams85:
> In order for Jint to easily leverage the new data slot for its selfish needs

Although this internal data slot is unused by Esprima currently, I'm not sure that we won't need it in the future. (As a matter of fact, it's not even completely unused because [`SyntaxToken` already make use of it](https://github.com/sebastienros/esprima-dotnet/blob/b8db670b207343da08350cb3e0437df18a4d6e01/src/Esprima/Ast/SyntaxToken.cs#L19-L25). I guess that's irrelevant to Jint though.)

If we decide to reserve the internal slot for Jint, we should at least document this fact in Esprima for future maintainers.

> I think it's easiest to open a public API. Exposing internals via attribute doesn't pan out very well as there are clashes with all the attributes that have been internalized in both projects, like DoesNotReturn etc.

If it's made available for Jint alone, this API seems fragile enough in itself, so I'd avoid to make it public at any cost. Couldn't Jint just reuse the internal attributes from Esprima without duplicating them?

Or maybe we should rethink the approach of attaching extra info to AST nodes? One that comes to my mind is [`ConditionalWeakTable<TKey,TValue>`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2?view=net-6.0). This, of course, doesn't help if we want to allow attaching arbitrary user data to nodes because that would require a dictionary anyway. The current solution is ok for that purpose. However, for attaching internal data without allocation, internal static `ConditionalWeakTable`s could be a viable solution.

How about some benchmarks?

lahma:
Maybe I should have elaborated about the *why* before I brought in my version is of *how* 🙂 Let's take this JS and think about esprima's `Identifier`:

```javascript
let x = {}; x["foo"]; x.constructor[Symbol.hasInstance];
````
A lot of identifiers, especially the `x` gets repeated. Jint sees here 7 distinct identifiers via AST (if I checked the AST right), though a layman might say there's `x` just once instead of three times as a concept. With these Jint needs to do different kind of lookups:

* `x` from lexical scope chain, considered a `string`, but Jint has `Key` that combines pre-calculated hash code + the string
* `x` like above and then `foo` that is a string, in Jint terms a wrapper called `JsString`
* like above, then another identifier for constructor and then known global symbol, a `JsSymbol`

When Jint is called with such script everything starts fresh, no memory of what's happened before. Jint tries to be smart and wraps AST again and in this case there's `JintIdentifier` that takes `Identifier` and pre-calculates as much as it can. This is silly to do every time same script is fed into engine and it should be cached on AST level (the node secret slot). Even sillier if the `x` is finally used only once.

So Jint needs to do fast `Identifier` to key (either `Key`, `JsString` or even `JsSymbol`) and then do lookup against chosen dictionary which uses same type of key `myDictionary[jintIdentifier.jsString]` or `lexicalScope[jintIdentifier.Key]` etc.

What I would not like to do is ask esprima to do another dictionary lookup (hashing + equals) in this critical path. So I would just read all cached data from `Identifier` instance's field, caching already happened in (yet to create) `Engine.PrepareScript(myString)` method.

Other use cases are like function hoisting and analysis (does it need arguments object etc), but such in total are more heavy and could live with yet another dictionary lookup to get state. If I'd have to do dictionary lookup to prepare for dictionary lookup it's an issue - I would need to keep the `JintIdentifier` and basically node's slot would give little comfort.

`OnNodeCreated` could be something like this when used in `PrepareScript`:

```c#
var dictionaryForIdentifiers;
if (node == identifier)
{
   node.internalSlot = dictionaryForIdentifiers.GetOrAdd(identifier.Name, () => new IdentifierCachingData(identifier.Name));
}
// other types
```

In Jin'ts case I would argue it's safe to say "If you need the slots in AST", parse one for yourself and give another one to Jint to mangle.

CharlieEriksen:
> Other use cases are like function hoisting and analysis (does it need arguments object etc)
FWIW, this is the part I'm really interested in, on top of the problem of resolving references. :)

> What I would not like to do is ask esprima to do another dictionary lookup (hashing + equals) in this critical path. So I would just read all cached data from Identifier instance's field, caching already happened in (yet to create) Engine.PrepareScript(myString) method.

Would this approach allow for other libraries to leverage some of that magic, like hoisting, without actually executing anything, just pure static analysis?

jogibear9988:
Why don't we use a Public Property like the "Tag" of each control in Windows Forms?

adams85:
> If I'd have to do dictionary lookup to prepare for dictionary lookup it's an issue - I would need to keep the JintIdentifier and basically node's slot would give little comfort.

Well, so `ConditionalWeakTable` and other dictionary-based solutions are out of the questions. Then we can do something like what I outlined in #338. This would allow Jint to attach extra information to nodes without interfering with Esprima internals:

```c#
var dictionaryForIdentifiers;
if (node == identifier)
{
   node.AdditionalData.StaticData = dictionaryForIdentifiers.GetOrAdd(identifier.Name, () => new IdentifierCachingData(identifier.Name));
}
// other types
```

The price is [an extra allocation](https://github.com/sebastienros/esprima-dotnet/pull/338/commits/8ea4f2b64d0d7ae275f87bce4f2a205f4d44f32c#diff-52c949261c9a5bfcd3f8e2af5165c8e06bda58c265a31f5bcb74bbab6819cd60R69) (of an `AdditionalDataContainer` instance) per identifier node. However, I don't think we can do much better if we want to provide a clean API which satisfies all the requirements and is economic implementation-wise at the same time.

Please review this alternative idea.

> Why don't we use a Public Property like the "Tag" of each control in Windows Forms?

We kinda do that currently (`AdditionalDataContainer` is a just wrapper around a single `object` field). A simple `Tag` property wouldn't cut it because we want to attach extra data internally and to allow end users to attach arbitrary data at the same time. We need to support both requirements with keeping the memory footprint of the AST as low as possible in cases where extra data is not needed.

jogibear9988:
Are there users wich need additional information on an Ast Node wich are not Jint and could not use a dictionary?

jogibear9988:
where do we need addtional data? was this with JSX?

adams85:
> Are there users wich need additional information on an Ast Node wich are not Jint and could not use a dictionary?

Users could always use a dictionary for associating data with AST nodes. Even Jint has done that so far.

However, not this is what guided the current design. Esprima itself needs to attach extra information to AST nodes in some circumstances:

> where do we need addtional data? was this with JSX?

E.g. when users want the parser to collect [tokens and/or comments](https://github.com/sebastienros/esprima-dotnet/blob/b8db670b207343da08350cb3e0437df18a4d6e01/src/Esprima/Ast/Expression.cs#L21).

This is easy in JS: you just add additional properties to the node objects at run-time. JS objects are essentially just dictionaries.

As opposed to that, C# is a strongly-typed language, you can only add properties at design-time. The problem with this is that if you use plain backing fields for these properties, they will consume memory even if their backing storage is unused. Which could increase the memory footprint of large ASTs significantly.

So we don't really have another way to workaround this than some lookup mechanism like dictionaries or `ConditionalWeakTable`s. We chose the former. And since we chose that, allowing also end users to use these dictionaries for their own purposes logically followed. It costs nothing more when we have those dictionaries in place anyway and provides a convenient API for end users to attach their additional data with nodes.

Of course, you could argue that we replaced a few large dictionaries with many small ones. However, keep in mind that no dictionaries are allocated until actually needed. Also, the memory slot of the dictionary reference can be used for storing some internal data without allocation up to that point.

My premise is (or was when I designed this) that in the majority of the cases no extra data is attached by the user. In opposite cases, where many nodes have additional data, it's obviously a worse solution in terms of memory consumption. Maybe even performance-wise but I can't tell much about that without benchmarking how lookups in large dictionaries compare with lookups in small ones.

I may put together another PR which drops dictionaries in favor of `ConditionalWeakTable`s so we can compare.

lahma:
Closing this as superseded by #338 , can always continue discussion if needed.

PR #338, opened 2022-09-08
Allow attaching user-defined data to nodes without dictionary lookup

adams85:
This is an alternative solution to the issue addressed by #337. It works without leaking internals through the public API (and at the same time makes it a bit cleaner) at the expense of a bit of extra allocation. (An `AdditionalDataContainer` instance is created when user-defined data is actually attached.)

### DIFF
--- a/src/Esprima/Ast/Expression.cs
+++ b/src/Esprima/Ast/Expression.cs
@@ -21,9 +21,9 @@ public abstract class Expression : StatementListItem, ISyntaxTreeRoot
     public IReadOnlyList<SyntaxToken>? Tokens
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get => (IReadOnlyList<SyntaxToken>?) AdditionalData[s_tokensAdditionalDataKey];
+        get => (IReadOnlyList<SyntaxToken>?) GetDynamicPropertyValue(TokensPropertyIndex);
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        set => AdditionalData[s_tokensAdditionalDataKey] = value;
+        set => SetDynamicPropertyValue(TokensPropertyIndex, value);
     }
 
     /// <summary>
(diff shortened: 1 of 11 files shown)


lahma (approved):
Somewhat more indirection to get the value, but I think this will suite my needs fine. Thanks for refining!

adams85:
After thinking over [the discussion in the other PR](https://github.com/sebastienros/esprima-dotnet/pull/337#issuecomment-1241091933), I came to the conclusion that using dictionaries per node for storing additional data/properties is maybe not the best idea after all. It enables a convenient API for end users but can also be pretty suboptimal in cases where users need to attach additional data to many nodes. I also didn't like the extra allocation required for attaching user data. I wanted an implementation which favors Jint's use case. I think finally I managed to come up with something like that:

With https://github.com/sebastienros/esprima-dotnet/pull/338/commits/3a44699ecec74f75bc2043d6afa94728976062b3 the dictionary-like API is no longer available but users can attach an arbitrary object via the `AssociatedData` property without penalty (i.e. without allocating extra objects). This is virtually the same as `Tag` in WinForms: it needs nothing more than [an extra type check](https://github.com/sebastienros/esprima-dotnet/pull/338/commits/3a44699ecec74f75bc2043d6afa94728976062b3#diff-52c949261c9a5bfcd3f8e2af5165c8e06bda58c265a31f5bcb74bbab6819cd60R19) compared to plain field access.

Now the Jint snippet would look like this:
```csharp
var dictionaryForIdentifiers;
if (node == identifier)
{
   node.AssociatedData = dictionaryForIdentifiers.GetOrAdd(identifier.Name, () => new IdentifierCachingData(identifier.Name));
}
// other types
```

So, I repeat: end users can attach a single object via `AssociatedData`, no more `node.SetAdditionalData(key, value)` or `node.AdditionalData[key] = value`. Are you ok with this? Pinging especially @CharlieEriksen as we [had a discussion about this](https://github.com/sebastienros/esprima-dotnet/issues/275#issuecomment-1163225918) earlier.

lahma (approved):
I like the API from Jint consumer side now, nice work!

lahma:
After we get this processed, I'd like to create another beta to NuGet for dogfooding.

lahma:
This has been published as [3.0.0-beta-7](https://www.nuget.org/packages/Esprima/3.0.0-beta-7)

lahma:
Jint has adopted the new version in https://github.com/sebastienros/jint/pull/1289

adams85:
I've also [updated my bundler to beta 7](https://github.com/adams85/bundling/commit/f7f92b203fa4f0a2b80a9a1e85074dd0f8b9017c). Everything seems to work fine. I only ran into one minor obstacle during the update process: it feels that introduction of `ScannerOptions` has made [this](https://github.com/adams85/bundling/commit/f7f92b203fa4f0a2b80a9a1e85074dd0f8b9017c#diff-92b345a131fd56a8f3975d6c4cc410aeef3886cbf7aeac38facb5ee471208087L59) more cumbersome than it should be. I'll see if I can do something about improving this API.

CharlieEriksen:
Updated my stuff to beta 7. So far, so good after adding in my own extensions for `GetAdditionalData` and `SetAdditionalData` :)

Issue #330, opened 2022-08-26
Missing ECMAScript 2020+ features 


adams85:
Readme/description says that the lib fully supports ECMAScript 2019. However, this statement looks a bit outdated in the light of recent additions: AFAICS, almost everything is covered up to ECMAScript 2022 already (albeit still buggy here and there). According to [this source](https://exploringjs.com/impatient-js/ch_new-javascript-features.html#new-in-es2020), here we are currently:

ECMAScript 2020
* [x] Dynamic imports via `import()`
* [x] Import metadata for the current module (`import.meta`)
* [x] Namespace re-exporting (`export * as ns from 'mod';`)
* [x] Optional chaining for property accesses and method calls (`value?.prop`)
* [x] Nullish coalescing operator (??) (`value ?? defaultValue`)
* [x] Bigints – arbitrary-precision integers
* [x] `globalThis` – access the global object that works both on browsers and server-side platforms (Note: as it seems that other parsers also treat this a simple identifier as opposed to some special syntax element like `ThisExpression`, we can consider this done.)

ECMAScript 2021:
* [x] Logical assignment operators (`a ||= b`, `a &&= b`, `a ??= b`)
* [x] Underscores (_) as separators in numeric and bigint literals (`123_456.789_012`, `6_000_000_000_000_000_000_000_000n`)

ECMAScript 2022:
* [x] Instance and static public fields:
    ```js
    class StaticPublicFieldClass {
      /* static */ identifierFieldKey = 1;
      /* static */ 'quoted field key' = 2;
      /* static */ [computedFieldKey] = 3;
    }
    ```
    Note: this looks implemented but buggy. Quoted and computed fields are parsed with error. @lahma Maybe you have something in the works related to this? (I mean #248)
* [x] Private fields, methods and accessors
    ```js
    class C1 {
      /* static */ #p_field = 0;
      /* static */ get #p_getter() {}
      /* static */ set #p_getter(v) {}
      /* static */ #p_method() {}
    }
    ```
    Note: again, implemented but buggy. The resulting AST contains `Identifier`s instead of `PrivateIdentifier`s. (Compare with AST Explorer output.)
* [x] Static initialization blocks
    ```js
    class C1 {
      static {}
    }
    ```
* [x] Private slot checks (`#privateSlot in obj`) Works but private identifier parsed incorrectly as `Identifier`. (Compare with AST Explorer output.)
* [x] Top-level await in modules
[...]

PR #343, opened 2022-10-02
Add support for RegEx d and v flags

lahma:
* https://github.com/tc39/proposal-regexp-match-indices
* https://github.com/tc39/proposal-regexp-v-flag

Small cleanup to parsing. I've validated the parsing on Jint side, unblocks regex issues. I'll release another beta to get this fix out.


adams85:
Nice improvement! 👍 I've updated #330 to keep track of implemented JS features.

Issue #344, opened 2022-10-07
Problem with the latest build on Xamarin/iOS


h15ter:
Hi Guys,

First of all I want to thank you for amazing projects you develop, Jint and Esprima worked well for me for many years.

I upgraded some older Xamarin app to the Jint@3.0.0-beta-41 and Esprima@3.0.0-beta-7 a few days ago, and it started to fail in runtime on iOS. If I downgrade to Jint@3.0.0-beta-39 and Esprima@3.0.0-beta-3, the problem goes away.

I didn't dig super-deep into it, but it looks like [this commit](https://github.com/sebastienros/esprima-dotnet/pull/329/files#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1) using open delegates for parser methods caching might be the root cause. 

I understand Xamarin/iOS is not 1st class citizen, but the library was fully compatible with it, and I really enjoyed that because it allowed me to use the same script execution engine across different platforms and runtime versions. Do you think it might be fixed in the library?

Thanks in advance!

```
System.ExecutionEngineException: Attempting to JIT compile method '(wrapper delegate-invoke) Esprima.Ast.Expression <Module>:invoke_callvirt_Expression_JavaScriptParser (Esprima.JavaScriptParser)' while running in aot-only mode. See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.

at Esprima.JavaScriptParser.ParseLeftHandSideExpressionAllowCall () <0x1061aa080 + 0x00694> in <0b544c2cdd8e4af3a57b8d45fca589f1#bef47091073c88a40e760639b4f843c2>:0 
at Esprima.JavaScriptParser.ParseUpdateExpression () <0x1061ac7e0 + 0x00287> in <0b544c2cdd8e4af3a57b8d45fca589f1#bef47091073c88a40e760639b4f843c2>:0 
at Esprima.JavaScriptParser.ParseUnaryExpression () <0x1061ad3b0 + 0x001ef> in <0b544c2cdd8e4af3a57b8d45fca589f1#bef47091073c88a40e760639b4f843c2>:0 
at Esprima.JavaScriptParser.ParseBinaryExpressionOperand () <0x1061adb00 + 0x001a3> in <0b544c2cdd8e4af3a57b8d45fca589f1#bef47091073c88a40e760639b4f843c2>:0 
[...]

PR #345, opened 2022-10-08
Fix AOT-related issues using function pointers

adams85:
This is a potential solution to #344 if you are ok with a bit of unsafe code ([C# 9 function pointers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers)). Of course, unsafe code should be avoided if possible, yet in this case we might consider it because besides solving the issue, it would also bring a minor performance improvement:

| Old |ParseProgram|angular-1.2.5|16.386 ms|0.0850 ms|4314.91 KB|
| **New** |	|	| **16.211 ms (-1%)** | **0.0765 ms** | **4314.91 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|2.187 ms|0.0073 ms|676.04 KB|
| **New** |	|	| **2.147 ms (-2%)** | **0.0069 ms** | **676.04 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|13.007 ms|0.0689 ms|3764.05 KB|
| **New** |	|	| **12.846 ms (-1%)** | **0.0645 ms** | **3764.05 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|20.247 ms|0.0386 ms|5797.76 KB|
| **New** |	|	| **20.086 ms (-1%)** | **0.0854 ms** | **5797.79 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|10.244 ms|0.0672 ms|3090.51 KB|
| **New** |	|	| **10.099 ms (-1%)** | **0.0071 ms** | **3090.51 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|1.866 ms|0.0089 ms|570.9 KB|
| **New** |	|	| **1.837 ms (-2%)** | **0.0065 ms** | **570.9 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|9.518 ms|0.0586 ms|2856.3 KB|
| **New** |	|	| **9.287 ms (-2%)** | **0.0549 ms** | **2856.3 KB (0%)** |

As for `MaxAssignmentDepth`, this approach has a minimal negative effect (considering the core parser): max. possible depth reduced by 3.

Are you ok with this or would you rather go the "safe" way?

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using Esprima.Ast;
@@ -10,7 +11,7 @@ namespace Esprima;
 /// <remarks>
 /// Use the <see cref="ParseScript" />, <see cref="ParseModule" /> or <see cref="ParseExpression(string)" /> methods to parse the JavaScript code.
 /// </remarks>
-public partial class JavaScriptParser
+public unsafe partial class JavaScriptParser
 {
     internal sealed class Context
     {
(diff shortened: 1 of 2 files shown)


adams85:
I made some further research and came to the conclusion that we'd better scrap this idea.

As mentioned in the feature documentation, function pointers are for static methods only, instance methods (or more precisely, the EXPLICITTHIS calling convention) are not supported directly. Although it would probably work in our case but would be considered a risky hack because we'd rely on unspecified behavior. ([This article](https://minidump.net/c-why-function-pointers-cant-be-used-on-instance-methods-8a99fc99b040) explains the details nicely.)

We could still use function pointers by converting the cached instance methods to static ones and adding an explicit this parameter but that would be too much pain for too little gain.

PR #347, opened 2022-10-08
 Revise usage of string operations which are culture-sensitive by default

Xicy:
When I tried to compile source code, sourceGenerator generate wrong content. This is caused by culture. I'm using in Tr-TR culture and it's try to convert 'i' to 'I'. This pr solve the issue when building.

Logs:

```log
Esprima.SourceGenerators.StringMatcherGenerator/StringMatchers.g.cs(159,1): error CS0246: 'ınternal' türü veya ad alanı adı bulunamadı (bir using yönergeniz veya derleme başvurunuz mu eksik?) 
```

```cs
"internal".ToLower() = "ınternal" //Problem
"internal".ToLowerInvariant() = "internal"
```

I → lowercase → ı
i → uppercase → İ

### DIFF
--- a/src/Esprima.SourceGenerators/StringMatcherGenerator.cs
+++ b/src/Esprima.SourceGenerators/StringMatcherGenerator.cs
@@ -111,7 +111,7 @@ public class StringMatcherGenerator : IIncrementalGenerator
             returnType = "string";
         }*/
 
-        var typeModifiers = method.ContainingSymbol.DeclaredAccessibility.ToString().ToLower();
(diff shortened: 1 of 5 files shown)


adams85:
Culture-sensitive behavior by default in (some of the) string operations is one of the few messed up design choices in .NET...

To make this fix complete, I suggest looking also for the following string operations and making sure that they are called using invariant culture: `Compare`, `StartsWith`, `EndsWith`, `IndexOf(string)`, `LastIndexOf(string)`

PR #348, opened 2022-10-15
Allow end users to directly instantiate Scanner using a ParserOptions instance

adams85:
After recent changes, no simple way exists for end users to convert a `ParserOptions` to a `ScannerOptions` instance when they want to create a `Scanner` with the same configuration. The proposed change would simplify use cases like [this](https://github.com/adams85/bundling/commit/f7f92b203fa4f0a2b80a9a1e85074dd0f8b9017c#diff-92b345a131fd56a8f3975d6c4cc410aeef3886cbf7aeac38facb5ee471208087L59).

### DIFF
--- a/src/Esprima/JavascriptParser.cs
+++ b/src/Esprima/JavascriptParser.cs
@@ -142,7 +142,7 @@ public partial class JavaScriptParser
         _maxAssignmentDepth = options.MaxAssignmentDepth;
         _onNodeCreated = options.OnNodeCreated;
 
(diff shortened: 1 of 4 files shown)


lahma (approved):
LGTM

adams85:
Thanks for the quick feedback!

lahma:
I've tagged this commit/build for release so we'll have more to dogfood again as 3.0.0-beta-9.

PR #349, opened 2022-11-09
Remove Data member from PropertyKind and make it non-Flags

lahma:
I could have sworn that this cleanup was already done earlier. `Data` is never used and `[Flags] `really makes no sense as you cannot have multiple active values.


lahma:
Ah it was cleaned in class PR, well now it actually has landed in main...

PR #353, opened 2022-12-25
Support accessor keyword in class properties

jogibear9988:


lahma:
I've updated the test suite so if you rebase and remove lines from `allow-list.txt` having `grammar-field-accessor` in their name (total of 4 lines) it should cover new cases that tests want to parse .

jogibear9988:
will do, at first i need to fix broken tests

jogibear9988:
@lahma please review.

I also needed to add some tests to the allow list, cause the generated _characterData seems not to be correct.

But I did not understand the regex to fit it

lahma (approved):
Generally looks good to me, the allow list shows that we are moving to the right direction. I'm willing to merge if ready, we probably need more cleanup to the class parsing logic later like you've discovered that it's a bit complex.

So should we merge?

jogibear9988:
for me it's ready, i'm done atm

lahma:
Thanks for going through the spec and implementation, always appreciated!

PR #356, opened 2022-12-26
Class and import related parsing fixes

### LINKED ISSUE #354: Wrong handling of private Identifierers
Try this:

        class aa {
            #aa = true;
        }

We do not create an instance of PrivateIdentifier. We create a normal node with "#" in the value

jogibear9988:
fixes #354


src/Esprima/JavascriptParser.cs
```diff
@@ -2074,6 +2078,10 @@ private UnaryExpression ParseBasicUnaryExpression()
         {
             TolerateError(Messages.StrictDelete);
         }
+        if (_context.Strict && unaryExpr.Operator == UnaryOperator.Delete && unaryExpr.Argument is MemberExpression m && m.ChildNodes.Last() is PrivateIdentifier)
```
adams85, line 2081: Enumeration (`m.ChildNodes.Last()`) looks unnecessary here. Wouldn't `m.Property` be enough?

jogibear9988: think so

lahma:
Looking really good so far, fixes a lot of parsing problems 👍🏻 Let us know when you think this is mergeable (needs a rebase at least).

jogibear9988:
im done

lahma (approved):

src/Esprima/Scanner.cs
```diff
@@ -2361,6 +2361,17 @@ internal string EscapeFailingRegex(string pattern)
             pattern = pattern.Replace("[^]", @"[\s\S]");
         }
 
+        if (pattern.Contains("--"))
+        {
+            pattern = pattern.Replace("--[", "-[");
+            var r = new Regex("--([^\\[\\]]*)");
```
sebastienros, line 2367: This should be a static/compiled regex

PR #361, opened 2022-12-28
Implement top-level await in modules feature

adams85:


lahma:
Can you also remove the feature exclude from test262 settings and make sure test cases pass?

adams85:
Sure, it's done. We have 4 failing test cases in test262 but those aren't related to this feature and will be fixed by #356. We can merge this PR after that.

lahma (approved):
Looking good!

PR #362, opened 2022-12-30
Additional improvements to class parsing

adams85:
Did a bit of yak shaving as I made an attempt at completing #330...

`ParseClassElement` got a complete overhaul: deep nesting got flattened by some `goto` usage, which, I know, has a bad reputation but I think this is the case where it makes things easier to follow.

Besides making the code more approachable, the PR also improves error reporting a bit and fixes a bunch of bugs. Further 161 cases could be removed from the allow-list. 8 cases returned though as it turned out they were false positives. (The previous implementation threw errors but not for the expected reasons.) These cases are related to `arguments` not being allowed in field initializers. E.g. `class C { static 'x' = () => arguments; }` AFAICS, it's not trivial to track this, so I'd just ignore this for now.


lahma (approved):
Nice cleanup, great work!

PR #364, opened 2022-12-30
Fix regexp parsing issues

adams85:
As I see, support for the [experimental `v` flag](https://2ality.com/2022/11/regexp-v-flag.html) has been enabled in regexp parsing, however I'm not sure that those regexps are adapted to .NET ones correctly. [This piece of code](https://github.com/sebastienros/esprima-dotnet/compare/fix/regex-parsing?expand=1#diff-5c649e707747c206cd95c805966b187aa0e3f0725f25f51aef684f8254ad0149L2366-L2368) looks like doing nothing else but make the corresponding test262 test cases pass. However, the adapted regexp doesn't seem to match the original one.

I think we should reconsider this. In my opinion no regexp would still be a better choice than a wrong one in those cases where the regexp can't be adapted correctly. Or, if we're ok with returning broken regexps, at least some warning should be presented to the consumer. What do you think?

---

Public API breaking changes:
* Replace the `Scanner.ParseRegex` instance method with a static method (`AdaptRegExp`).
* Remove the `Scanner.ParseRegexOptions` method as it is not intended for end users.
* Deprecate the `ParserOptions.AdaptRegexp`/`ScannerOptions.AdaptRegexp` properties in favor of `RegExpParseMode`. `AdaptRegexp` will be removed from the final v3.0 release.
* Change the `ParserOptions.ScannerOptions` property into a method (`GetScannerOptions`) to prevent it from appearing in intellisense. (Probably I'm the sole consumer of this API, so this shouldn't break much code.)


lahma:
Ideally we would of course transform the ones we can. But if it's known to be broken maybe throwing an exception is the right choice instead of getting wrong results from script evaluation. NilJS does [some unicode conversions](https://github.com/nilproject/NiL.JS/blob/develop/NiL.JS/BaseLibrary/RegExp.cs#L162), but it doesn't have anything for the `v` flag.

adams85:
Thank you. 😃 I can't promise though that this will completely solve the issues around regexes. It's pretty hard indeed to get it right because of the sheer amount of quirks and edge cases. But hopefully it will improve the situation somewhat. Besides getting rid of the quick&dirty patches and fixing as many bugs as I can, the main goal is to refactor this part so we can maintain it more easily in the future.

sebastienros:
One day I will disable force-push so that I can actually see what was done in the last commit ;) (don't worry just kidding). Story is that I receive the notification of the update, but this link shows an empty changeset because it's a force-push so I can't see what was changed, it's like a brand new PR. I personally usually do that too, but only when rebasing. And I then squash during the merge to delete the history.

adams85:
Sorry for spamming you with notifications. You can just ignore these for now since this is a draft. I'm just pushing changes to back up local work as I've invested a lot in this one and I'd tear my hair out if I lost it because of a hardware failure... :D I wasn't aware though that others receive notifications of changes pushed to draft PRs. Of course, I'll quit doing this if it's annoying to you.

adams85:
Final force push, PR is ready 🎉 

The implementation is far from perfect but I think it's pretty close to what can be achieved using the approach of adapting JS regex patterns for the .NET regex engine to execute.

I worked hard to make the behavior of the adapted patterns as close as possible to the original one but there are limits to this due to the implementation differences of the two regex engines. We can work around many of them but unfortunately not all of them.

Known issues:
* Case-insensitive matching [won't always yield the same results](https://github.com/sebastienros/esprima-dotnet/blob/22eab845300564fcf756aab548ff635c58c4cd18/src/Esprima/Scanner.RegexParser.cs#L91). Implementing a workaround for this issue would be extremely hard, if not impossible.
*  Self-referencing capturing groups like `/((a+)(\1) ?)+/` may not produce the exact same captures. [`RegexOptions.ECMAScript` is supposed to cover this](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#ecmascript-matching-behavior), however even the MSDN example don't produce the same matches. (BTW, `RegexOptions.ECMAScript` is kinda a false promise, it helps very little overall. MS didn't even get `.` or `\s` matching right...)
* Similarily, repeated nested groups like `/((a+)?(b+)?(c))*/` may produce different captures for the groups. ([JS has an overwrite behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Capturing_group#description) while .NET don't).
* .NET treats forward references differently than JS, so it's not possible to convert patterns like `\1(\w)` reliably. (We could make some patterns work by rewriting them to something like `(?:)(\w)` but I found cases where even this wouldn't work.)
* If there are multiple syntax errors, our parser won't report the exact same error as V8 in some cases. (This might be improved but I don't think it would be worth the effort.)
* Unicode mode issues:
  * There could be false positive empty string matches in the middle of surrogate pairs. Patterns as simple as `/a?/` have this issue when the input string contains astral Unicode chars. I didn't find a workaround for this but I'm pretty sure that, if it's exists, it's very complicated. This is a **big problem but can be mitigated somewhat using a bit of "post-processing"**, that is, filtering out the false positive matches after evaluation like it's done [here](https://github.com/sebastienros/esprima-dotnet/blob/22eab845300564fcf756aab548ff635c58c4cd18/test/Esprima.Tests/RegExpTests.cs#L296). I'm afraid we can't do anything better than this until .NET adds the options to treat the input string as Unicode code points.
  * Support for Unicode property escapes is pretty limited (see the explanation [here](https://github.com/sebastienros/esprima-dotnet/blob/22eab845300564fcf756aab548ff635c58c4cd18/src/Esprima/Scanner.RegexParser.Unicode.cs#L831)). Currently, we can only adapt General Category expressions. But even this is not perfect as the result will depend the Unicode version included in the specific .NET runtime which runs the code.

So, to sum it up, I think we can get about this far by the current approach. Legacy pattern conversion is pretty solid apart from the minor issues listed above. However, we still have nothing for flag v but TBH I don't have much hope for that since even flag u support is partial and quirky.

If we wanted a fully compliant solution, we'd need to implement a regex engine based on the JS specs. Which, of course, would be a HUGE undertaking. I don't see that happening in the near future, however if someone wants to give it a shot, they can reuse a lot of the code from this PR for the parser part.

adams85:
As a first step, @lahma, could you do some tests with Jint? You may expect a BC because I changed the method for standalone regex parsing in `Scanner`: https://github.com/sebastienros/esprima-dotnet/blob/22eab845300564fcf756aab548ff635c58c4cd18/src/Esprima/Scanner.cs#L1786

lahma:
Will do, probably tomorrow the latest.

lahma:
I did a quick test with Jint and seems that there are tests failing now with adapted usage (using Debug mode with asserts). Errors like:

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Esprima.Scanner.RegexParser.LegacyMode.AdjustEscapeSequence(AdjustPatternContext& context, RegexParser& parser) in D:\Work\esprima-dotnet\src\Esprima\Scanner.RegexParser.Legacy.cs:line 143
   at Esprima.Scanner.RegexParser.AdjustPattern[TMode](TMode mode, ArrayList`1& capturingGroups, Dictionary`2 capturingGroupNames) in D:\Work\esprima-dotnet\src\Esprima\Scanner.RegexParser.cs:line 568
```

```
Esprima.ParserException: Line 1: Invalid regular expression: Nothing to repeat
   at Esprima.Scanner.ThrowUnexpectedToken(String message) in D:\Work\esprima-dotnet\src\Esprima\Scanner.cs:line 187
   at Esprima.Scanner.RegexParser.AdjustPattern[TMode](TMode mode, ArrayList`1& capturingGroups, Dictionary`2 capturingGroupNames) in D:\Work\esprima-dotnet\src\Esprima\Scanner.RegexParser.cs:line 532
```

I'm not sure if I can quickly suggest any fixes, but if you take Jint `main` branch and remove NuGet package reference to Esprima and add solution reference to local project + Jint reference to that new project you should be able to see the problems. 

I don't want to depress you - the improvements look great and in a way also sad that so much adaptation is needed to work around .NET regex "ecmascript" mode problems. Awesome work as always 👍🏻

lahma:
Also API thought, should `Scanner.ParseRegex` not ever return nullable and have another `Try` version. I know the tolerant mode can be a bit ambiguous in this context but just a thought.

adams85:
First of all, thanks for the quick response.

> I'm not sure if I can quickly suggest any fixes, but if you take Jint main branch and remove NuGet package reference to Esprima and add solution reference to local project + Jint reference to that new project you should be able to see the problems.

Did this, however `dotnet test` reported only 6 failing tests in the test262 suite for me, all of them failing with `NullReferenceException` caused by the current error reporting behavior of the regex parser + the test executor, which parses in tolerant mode (more on this below).

The issue hit by these tests is this: ".NET treats forward references differently than JS, so it's not possible to convert patterns like \1(\w) reliably." So we need to exclude them from the test suite.

However, I got no `IndexOutOfRangeException` and `Nothing to repeat` errors. Could you tell me which test cases produced such errors for you?

> Also API thought, should Scanner.ParseRegex not ever return nullable and have another Try version. I know the tolerant mode can be a bit ambiguous in this context but just a thought.

That's a good point and I'm absolutely open for changes in this area. But first, let me explain the current behavior:

The new regex parser does complete syntactic validation and it throws non-tolerable exceptions when encountering a syntax error. However, when the pattern is syntactically correct but it can't adapt it to an equivalent .NET pattern because of the known issues I mentioned above, it's a tolerable error. In this cases `null` is returned when the error is tolerated (and it is by the default error handler). In other words, when it's the parser's fault that it can't produce a regex, it's a tolerable error, when it's the user's (as they gave the parser an invalid pattern), it's an intolerable error.

When I implemented the parser, this seemed to be a good idea. However, as I'm thinking about it, I'm not sure any more... Should be a syntax error in a regex an intolerable error in the first place? When you parse a script and you have an invalid regex in that, would it be better if parsing could continue? I think we should decide this first. What do you think?

lahma:
Been a bit busy with life, sorry for the delays...

> However, I got no `IndexOutOfRangeException` and `Nothing to repeat` errors. Could you tell me which test cases produced such errors for you?

Totally my bad, I did my testing in a hurry and expected that `null` result meant that parsing failed *and no exception would be thrown anymore*. Sorry and thanks for clarifying.

> > Also API thought, should Scanner.ParseRegex not ever return nullable and have another Try version. I know the tolerant mode can be a bit ambiguous in this context but just a thought.

> That's a good point and I'm absolutely open for changes in this area. But first, let me explain the current behavior:

> The new regex parser does complete syntactic validation and it throws non-tolerable exceptions when encountering a syntax error. However, when the pattern is syntactically correct but it can't be adapted to an equivalent .NET pattern because of the known issues I mentioned above, it's a tolerable error. In this cases `null` is returned when the error is tolerated (and it is by the default error handler). In other words, when it's the parser's fault that it can't produce a regex, it's a tolerable error, when it's the user's (as they gave the parser an invalid pattern), it's a non-tolerable error.

> When I implemented the parser, this seemed to be a good idea. However, as I'm thinking about it, I'm not sure any more... Should be a syntax error in a regex an intolerable error in the first place? When you parse a script and you have an invalid regex in that, would it be better if parsing could continue? I think we should decide this first. What do you think?

I think you original reasoning is quite sound. From interpreter perspective, it shouldn't be a problem if there's a regex that we cannot run *as long as we don't enter the code path*. The problematic use case might be imaginary validation service that would run against input and every once in a blue moon it would error because specific `if` or something would be entered and regex would be tried to be invoked - "I'm sorry but we never understood the regex on .NET side even though you've written perfect ECMAScript regex here - now it's here on your face".

So I'm afraid I don't have any definitive clear picture what would be the correct way to handle. There are pros and cons for every approach I'd guess.

adams85:
> I think you original reasoning is quite sound. From interpreter perspective, it shouldn't be a problem if there's a regex that we cannot run as long as we don't enter the code path. The problematic use case might be imaginary validation service that would run against input and every once in a blue moon it would error because specific if or something would be entered and regex would be tried to be invoked - "I'm sorry but we never understood the regex on .NET side even though you've written perfect ECMAScript regex here - now it's here on your face".

According to this, the current error reporting mechanism looks good enough. When the input has regexes with syntax errors, you get `ParserException`. In other cases, you have a choice: either you get a `ParserException` for correct but not adaptable regexes, or, by enabling tolerant mode, you can parse the script and decide later what to do with the problematic regexes.

I think the same behavior is good enough for parsing a single regex:

```cs
/// <summary>
/// Parses an ECMAScript regular expression and tries to create a <see cref="Regex"/> instance with an equivalent behavior.
/// </summary>
/// <returns>
/// The equivalent <see cref="Regex"/> if the conversion was possible, otherwise <see langword="null"/>.
/// </returns>
/// <exception cref="ParserException">
/// <paramref name="pattern"/> is an invalid regex pattern or, when <paramref name="tolerant"/> is <see langword="false"/>,
/// cannot be converted to an equivalent <see cref="Regex"/>.
/// </exception>
public static Regex? ParseRegex(string pattern, string flags, TimeSpan? matchTimeout = null, bool tolerant = true, ErrorHandler? errorHandler = null)
```

I'd remove `errorHandler` though because that doesn't seem to make much sense in this case. And I'd probably rename the method to `AdaptRegex` because that would describe the purpose better. Would you ok with this?

Regarding this topic, I have one more potential improvement in mind: we may allow syntax validation without adapting the regex, which may be a pretty expensive operation. So I'd introduce an enum like this:

```cs
public enum RegexParseMode
{
    Skip,
    Validate,
    Adapt,
}
```

`Skip` would correspond to the current `AdaptRegexp = false`, while `Adapt` to the `AdaptRegexp = true` setting. And there would be a third option, which would only validate but wouldn't create a .NET regex. Also, [probably we should change the default](https://github.com/sebastienros/esprima-dotnet/issues/371#issue-1556529137) to `Validate`.

Oh, and while we're at it, so far interpreted .NET regexes have been created. Probably it would make sense to add an option for compiled regexes (`RegexOptions.Compiled`). For example, could look like this:

```cs
public enum RegexParseMode
{
    Skip,
    Validate,
    AdaptToInterpreted,
    AdaptToCompiled,
}
```

What's your opinion on these ideas?

adams85:
After consulting with my pillow, this would be my final version of the single regex parser API:

```csharp
/// <summary>
/// Parses an ECMAScript regular expression and tries to create a <see cref="Regex"/> instance with the equivalent behavior.
/// </summary>
/// <returns>
/// The equivalent <see cref="Regex"/> if the conversion was possible,
/// otherwise <see langword="null"/> (unless <paramref name="throwIfNotAdaptable"/> is <see langword="true"/>).
/// </returns>
/// <exception cref="ParserException">
/// <paramref name="pattern"/> is an invalid regex pattern or cannot be converted
/// to an equivalent <see cref="Regex"/> (if <paramref name="throwIfNotAdaptable"/> is <see langword="true"/>).
/// </exception>
public static Regex? AdaptRegExp(string pattern, string flags, TimeSpan? matchTimeout = null, bool throwIfNotAdaptable = false)
```

I wouldn't split this into an `AdaptRegex` / `TryAdaptRegex` pair because then it would be harder to differentiate the user problem (syntax error) from the parser problem (not adaptable pattern).

(A note about naming: 'regex' is spelled in multiple ways currently. I think it'd be preferable to use `RegExp` when referring to JS and `Regex` when refererring to .NET regexes.)

lahma:
> After consulting with my pillow

😄 

> ```cs
> /// <summary>
> /// Parses an ECMAScript regular expression and tries to create a <see cref="Regex"/> instance with the equivalent behavior.
> /// </summary>
> /// <returns>
> /// The equivalent <see cref="Regex"/> if the conversion was possible,
> /// otherwise <see langword="null"/> (unless <paramref name="throwIfNotAdaptable"/> is <see langword="true"/>).
> /// </returns>
> /// <exception cref="ParserException">
> /// <paramref name="pattern"/> is an invalid regex pattern or cannot be converted
> /// to an equivalent <see cref="Regex"/> (if <paramref name="throwIfNotAdaptable"/> is <see langword="true"/>).
> /// </exception>
> public static Regex? AdaptRegExp(string pattern, string flags, TimeSpan? matchTimeout = null, bool compiled = false, bool throwIfNotAdaptable = false)
> ```
> 
> I wouldn't split this into an `AdaptRegex` / `TryAdaptRegex` pair because then it would be harder to differentiate the user problem (syntax error) from the parser problem (not adaptable pattern).
> 
> (A note about naming: 'regex' is spelled in multiple ways currently. I think it'd be preferable to use `RegExp` when referring to JS and `Regex` when refererring to .NET regexes in our APIs.)

This looks good to me, really appreciate the work you've done here.

adams85:
Life happened here too but at last I completed this one... I know it's quite a bit but I wouldn't like to merge this without a fair peer review. Could any of you do it?

src/Directory.Build.props
```diff
@@ -5,7 +5,7 @@
     <Copyright>Sebastien Ros</Copyright>
     <Authors>Sebastien Ros</Authors>
 
-    <LangVersion>latest</LangVersion>
+    <LangVersion>11</LangVersion>
```
lahma, line 8: maybe could add comment why needed so can be later `latest` when the needed feature is in latest default SDK

src/Esprima/JsxParser.cs
```diff
@@ -394,7 +394,7 @@ private string ScanXHTMLEntity(char quote)
             else if (hex && str.Length > 2)
             {
                 result =
-                    ((char) int.Parse(str.Substring(2), System.Globalization.NumberStyles.HexNumber)).ToString();
+                    ((char) int.Parse(str.Substring(2), System.Globalization.NumberStyles.AllowHexSpecifier)).ToString();
```
lahma, line 397: probably not needed for this PR, but we also have `HAS_SPAN_PARSE` defined when can use span based parsing, some linting tools will also probably warn about not using invariant culture

lahma (commented):
I went through the code on high level as I believe that you have studied the problem area and I can only agree with the outcome. A lot of great work here and seems that things are going towards the better (sad to have all the code needed  as Regex ECMAScript mode doesn't handle things as it should).

Left couple comments but IMO you can merge when you feel it's ready.

src/Directory.Build.props
adams85: As a matter of fact, this change is not necessary, I just like to be explicit about the C# version required for building the project. To my understanding, the meaning of `latest` depends on the .NET SDK installed, which just doesn't seem like a good idea because one may have an older SDK installed, which resolves `latest` to an older C# version than required. (E.g. the highest target we have is .NET 6 but that only includes C# 10.) Please correct me if I'm wrong or if you'd just like to have `latest` and I'll revert it.

src/Esprima/JsxParser.cs
adams85: I'll take care of this and similar minor issues in a separate PR.

src/Esprima/ParserOptions.cs
```diff
@@ -34,17 +37,27 @@ public record class ParserOptions
...
     /// Gets or sets the <see cref="ErrorHandler"/> to use, defaults to <see cref="ErrorHandler.Default"/>.
     /// </summary>
-    public ErrorHandler ErrorHandler { get => ScannerOptions._errorHandler; init => ScannerOptions._errorHandler = value; }
+    public ErrorHandler ErrorHandler { get => _scannerOptions._errorHandler; init => _scannerOptions._errorHandler = value; }
 
     /// <summary>
     /// Gets or sets whether the Regular Expression syntax should be converted to a .NET compatible one, defaults to <see langword="true"/>.
     /// </summary>
-    public bool AdaptRegexp { get => ScannerOptions._adaptRegexp; init => ScannerOptions._adaptRegexp = value; }
+    [Obsolete($"This property is planned to be removed from the next stable version. Please use the {nameof(RegExpParseMode)} property instead.")]
```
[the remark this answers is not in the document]
adams85: This one is not entirely clear to me. Instead of readme, didn't you mean release notes? That would make more sense. However, I'd also keep the obsolete attribute so early adopters can prepare their code for this change.

src/Esprima/Scanner.RegExpParser.Unicode.cs
```diff
@@ -0,0 +1,960 @@
...
+                {
+                    case 'd': // [0-9]
+                        set.Add(new CodePointRange(0x0030, 0x0039));
+                        break;
+                    case 'D': // [^0-9]
+                        set.Add(new CodePointRange(0x0000, 0x002F));
+                        set.Add(new CodePointRange(0x003A, Character.UnicodeLastCodePoint));
+                        break;
+                    case 's': // [\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]
+                        set.Add(new CodePointRange(0x0009, 0x000D));
```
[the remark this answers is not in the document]
adams85: Good point! I'll check if it can be improved.

adams85:
Thank you for taking a look so quickly, @lahma. And also for the useful suggestions. Hopefully, there are no major gotchas left because of the comprehensive set of tests (including the ~2200 test cases added which, besides syntax issues, also verify the behavior of the converted patterns).

So this should indeed improve the quality of the parser, by bringing it closer to the specs. However, this conversion approach won't bring us much further, I'm afraid. One thing could help though: if MS implemented an evaluation mode in the .NET regex engine which treats the input as Unicode code points (as flag u and v does), that would allow us to significantly improve the conversion of unicode patterns and to implement a fair support for unicode sets (flag v). So it may be worth lobbying MS (e.g. [here](https://github.com/dotnet/runtime/issues/79982)) to do that. (At least, I don't expect them to implement a fully compliant ECMAScript mode.)

src/Directory.Build.props
lahma: OK, maybe the explicit version will make things more.. explicit 👍🏻 I've just used to having latest SDKs and (hoping to have latest) CI tools.

src/Esprima/ParserOptions.cs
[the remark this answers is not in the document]
lahma: Yes, I meant any documentation that would accompany the final 3.x release. I was just wondering how many RCs there will be before the next stable - but would make sense to have at least one to to prepare people who live on the edge - I just wouldn't want obsolete attributes in final 3.0 as we've been doing so much cleanup already and it's again another TODO for 4.x 😉 .

lahma:
I added some replies, thank you for the discussion - all seems to make sense for me at least 👍🏻 

> Thank you for taking a look so quickly, @lahma. And also for the useful suggestions. Hopefully, there are no major gotchas left because of the comprehensive set of tests (including the ~2200 test cases added which, besides syntax issues, also verify the behavior of the converted patterns).

👍🏻 

> So this should indeed improve the quality of the parser, by bringing it closer to the specs. However, this conversion approach won't bring us much further, I'm afraid. One thing could help though: if MS implemented an evaluation mode in the .NET regex engine which treated the input as Unicode code points (as flag u and v does), that would allow us to significantly improve the conversion of unicode patterns and to implement a fair support for unicode sets (flag v). So it may be worth lobbying MS (e.g. [here](https://github.com/dotnet/runtime/issues/79982)) to do that. (At least, I don't expect them to implement a fully compliant ECMAScript mode.)

Yes it's a sad state of affairs for the EcmaScript mode - I guess full framework release cadence + fear of breaking changes made this quite impossible. Probably hard to fix in modern .NET now too.

lahma (approved):

src/Esprima/ParserOptions.cs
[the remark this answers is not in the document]
adams85: >  I just wouldn't want obsolete attributes in final 3.0
Fully agreed. I also have it in mind to remove the deprecated stuff from the final v3.x.

>  I was just wondering how many RCs there will be before the next stable
Hopefully, we won't need more than one or two RCs. AFAICS, [this one](https://github.com/sebastienros/esprima-dotnet/pull/381) can be merged soon. Apart from that, I have some minor polishing left.

BTW, I added a BC section to the top comment to help us with writing the release notes.

adams85:
As for me, PR is ready to merge. I'll wait a few days though before merging so others can put in their 2 cents if they want to.

PR #372, opened 2023-01-27
Generate visitation boilerplate code

adams85:
There is a bunch of boilerplate code related to visitation/enumeration of AST nodes which is time consuming and error-prone to maintain manually. A part of this (state machines of child node enumerations) is already auto-generated. This proposal, however, would take this effort to the next level. The following items would be generated based on the annotations of AST nodes (`VisitableNodeAttribute`) and visitors (`AutoGeneratedAstVisitorAttribute`):
* `NextChildNode` methods of annotated AST nodes (along with the related state machine helpers)
* `Accept` methods of annotated AST nodes
* `UpdateWith` methods of annotated AST nodes
* `VisitXXX` methods of annotated AST visitors (`AstVisitor`, `AstRewriter`, `JsxAstVisitor`, `JsxAstRewriter` + `IJsxVisitor`)

This would also pave the way for adding a generic visitor if we decided so.

One thing which is certainly still missing: auto-generation of `AstVisitorEventSource`/`JsxAstVisitorEventSource`. I haven't done this because I'm not sure that we need this one in the library at all. I can't really see the value in this visitor and don't know other parsers which include something like this either. So I think we may consider removing it (which would be a BC, of course). So, first of all, I'm pinging @atifaziz, who added it to the lib, if he'd be ok with removing? I'd like other collaborators to add their opinions about this too.


adams85:
To move things forward, could I get a review on this? In theory, no logic or public API has been changed, just a bunch of hand-written code has been replaced with auto-generated code, which can be viewed [here](https://github.com/sebastienros/esprima-dotnet/tree/0229894cb486fcc49f96076fa8a43a9fd9503afb/test/Esprima.Tests.SourceGenerators/Snapshots).

As for `AstVisitorEventSource`/`JsxAstVisitorEventSource`, we can postpone the decision on whether to remove or keep them, so I propose marking them as obsolete and let's see how is that received by the public. If no-one is missing them, I think we can remove them from the final 3.0 release.

lahma (approved):
LGTM, always nice to remove the repetition even tough we get a pinch of magic behind the scenes in return 👍🏻

atifaziz:
> I can't really see the value in this visitor and don't know other parsers which include something like this either. So I think we may consider removing it (which would be a BC, of course). So, first of all, I'm pinging @atifaziz, who added it to the lib, if he'd be ok with removing? I'd like other collaborators to add their opinions about this too.

Thanks for the ping and checking in. The value proposition of `AstVisitorEventSource` is provided in the [opening comment](https://github.com/sebastienros/esprima-dotnet/pull/73#issue-398399195) of PR #73:

> This avoids the ceremony of having to subclass `AstVisitor` for ad-hoc cases. Just create `AstVisitorEventSource` and subscribe to the events of interest.

Hopefully that still makes sense. If not, the following could be a fair call too:

> so I propose marking them as obsolete and let's see how is that received by the public. If no-one is missing them, I think we can remove them from the final 3.0 release.

adams85:
@lahma 
> LGTM, always nice to remove the repetition even tough we get a pinch of magic behind the scenes in return 👍🏻

Thanks for doing the review! I believe that we'll gain more from this little source generation sorcery (especially, when adding new stuff), than we'll lose in explicitness. Luckily, we'll always have the option to revert to code all this boilerplate manually. 😄

@atifaziz

Thank you also for popping in to add your thoughts on my proposal. Now it's clear what purpose this visitor in question serves. I'm just not sure that it's worth maintaining if it has no real use beyond ad-hoc cases (at least, I can't really imagine such use cases as I mentioned earlier). For some ad-hoc tinkering we also have enumeration, so the AST can be traversed without using a visitor at all.

But let's approach the question pragmatically. Are you actively using this feature, thus, do you need it included in the library in the future? If so, we'll keep it (and I'll write the generator for it). If not, let's deprecate it and see if there are others who need it.

atifaziz:
> I'm just not sure that it's worth maintaining if it has no real use beyond ad-hoc cases (at least, I can't really imagine such use cases as I mentioned earlier).

> For some ad-hoc tinkering we also have enumeration, the AST can be traversed without using a visitor at all.

I'm afraid I haven't followed the project enhancements closely so I don't know about the new methods that could make `AstVisitorEventSource` & co redundant, but do you mean traversal capabilities as I added in PR #77?

> Are you actively using this feature, …

No, but I can't say if code where I used it is still running in production or not.

> …thus, do you need it included in the library in the future?

I can't say what the distant future holds, but nothing in the foreseeable future. 🙂

> If not, let's deprecate it and see if there are others who need it.

Yep, fair call.

If I return to this space, I'd be happy to get up to speed on what's there and debate reviving `AstVisitorEventSource` if I need it and can still make a strong case for it.

adams85:
> I'm afraid I haven't followed the project enhancements closely so I don't know about the new methods that could make AstVisitorEventSource & co redundant, but do you mean traversal capabilities as I added in PR https://github.com/sebastienros/esprima-dotnet/pull/77?

Yep, essentially I had `DescendantNodes` in mind. I know that it does pre-order traversal while `AstVisitorEventSource` can do post-order as well. So it's true that it doesn't completely makes it redundant in this aspect. However, a post-order version of `DescendantNodes` is also possible. But again, the question is that it's considered such base functionality which a plain parser library should provide out of the box. I'm leaning towards not really. (BTW, the main enhancements in the area of enumeration was performance-related which, of course, doesn't make much difference in ad-hoc use cases.)

> Yep, fair call.

Alright, let's do this then as I proposed.

> If I return to this space, I'd be happy to get up to speed on what's there and debate reviving AstVisitorEventSource if I need it and can still make a strong case for it.

Looking at your PRs I see that you've added much value to the project already, it would be great to have you on board again. So, I'm looking forward to future discussion about `AstVisitorEventSource` or any other topic which catch your imagination! ;)

PR #375, opened 2023-02-19
Fix building of release artifacts in GitHub Actions

adams85:
Looks like we've been bitten by [this BC recently introduced in the .NET 7 SDK](https://github.com/dotnet/sdk/issues/30625).

Also, we've been polluting NuGet & MyGet with [our internal source generator project](https://www.nuget.org/packages/Esprima.SourceGenerators) for a while...

This PR attempts to solve both issues.


lahma (approved):
Looks good, I had to do [similar fix](https://github.com/sebastienros/jint/pull/1439) to Jint just yesterday. Quite the breaking change even though the consider it change for the better.

PR #377, opened 2023-02-19
Remove import assertions feature

adams85:
As the [import assertions feature](https://github.com/tc39/proposal-import-assertions) has been [demoted from Stage 3 to Stage 2](https://docs.google.com/presentation/d/1HbWhyo4tSnpv4vMZqCa2YQvi_mKdpDi4JWWBtSUQqQY/edit#slide=id.p) and the syntax may change in the future, I propose to remove this feature until finalized.

If the syntax won't change, we'll be able to put the feature back pretty easily by reverting these changes.

Please let me know if you disagree. If you think that this feature should be still included, I'm ok with that.


lahma (approved):
LGTM, shame to remove for the time being, but wise.

lahma:
@adams85 is this the bits that would be blocking releasing RC1, or maybe something more?

adams85:
Nope. Sorry, I forgot to merge this.

I have no further BCs to the public API in mind and AFAICS there are no issues which would need a BC either. So, it looks we're good for a rc1!

Issue #380, opened 2023-04-08
Import Assertions (aka Import Atributes) are back at stage 3 so we should add them again


jogibear9988:
see: https://github.com/tc39/proposal-import-attributes

jogibear9988:
we also need to rename the property "assertions" to "attributes", see : https://github.com/estree/estree/blob/master/experimental/import-attributes.md

adams85:
[PR](https://github.com/sebastienros/esprima-dotnet/issues/380) is ready but let's not merge it yet. I wrote about the why in the PR description.

PR #381, opened 2023-04-09
Add import attributes feature

adams85:
Re-adds the [experimental Import Attributes feature](https://github.com/tc39/proposal-import-attributes) (formerly known as Import Assertions).

See also:
* https://github.com/sebastienros/esprima-dotnet/issues/380
* https://github.com/sebastienros/esprima-dotnet/pull/377

The proposal seems to be WIP ("reached conditional consensus"), so I suggest we don't merge this for now but wait until they finish [this issue](https://github.com/tc39/proposal-import-attributes/issues/137) or, rather, until they implement it in V8 because of [this](https://github.com/sebastienros/esprima-dotnet/compare/feature/import-attributes?expand=1#diff-03bb493a0c607cc31255cf2fdb2dbdd87769675a7a5b17c88be39296f44b903cR25-R26).


adams85:
Looks like [the proposal is as final as it can get](https://github.com/microsoft/TypeScript/issues/53656#issuecomment-1635933314) now, so I resolved the merge conflicts and we can merge this as well. Any objections?

test/Esprima.Tests.Test262/Test262Harness.settings.json
```diff
@@ -5,8 +5,8 @@
   "Namespace": "Esprima.Tests.Test262",
   "Parallel": true,
   "ExcludedFeatures": [
-    "regexp-v-flag",
-    "import-assertions"
+    "import-assertions",
```
lahma, line 8: should this exclude be removed if this is the feature that should be parsed? might also be worth use the latest commit sha from test262 repo for correct input

adams85: The `import-assertions` feature is for the old version of the proposal. We doesn't support that, so it should be excluded.

Unfortunately, [the updated tests](https://github.com/tc39/test262/pull/3843) haven't landed in the test262 repo yet.

What do you think, how should we proceed? Shall we wait with merging our PR until the updated tests get merged in test262?

lahma: I think merge is fine as we can always add tests later/enable feature and fix possible regressions 👍🏻

adams85: Ok, then I'll go ahead. Also, I've subscribed to the test262 PR and will check those tests as soon as they become available.

Issue #382, opened 2023-04-09
Inconsistencies with ESTree


adams85:
We have a statement in the readme saying:

> Sensible [syntax tree format](https://github.com/estree/estree/blob/master/es5.md) as standardized by [ESTree project](https://github.com/estree/estree)

The first part is true but I'm afraid we can't claim that we strictly adhere to the ESTree standard at the moment as there are some differences between ESTree and our model. Personally, I'm ok with most of these differences as they are more or less forced by C#'s type system or in some cases the design choices of ESTree just look suboptimal from a practical standpoint. E.g in the standard model
* `LogicalExpression` doesn't extend `BinaryExpression`,
* `PrivateIdentifier` extends `Node` whereas `Identifer` extends `Expression`,
* `Super` extends `Node` whereas `ThisExpression` extends `Expression`,
* etc (I can list a few more if you're interested).

There may be theoretical reasons for these design choices but I'm quite happy with our AST model because it has proven to work pretty well in practice. However, I suggest changing the aforementioned statement to something like "Sensible [syntax tree format](https://github.com/estree/estree/blob/master/es5.md) **which follows the standard** established by the [ESTree project](https://github.com/estree/estree)".

I found only one issue (a naming inconsistency) which still contradicts this claim somewhat: [the standard name of dynamic import nodes is `ImportExpression`](https://github.com/estree/estree/blob/70f89cddd0bb3d9842847314d34c32f315b9336d/es2020.md#importexpression) instead of `Import`. This is the other change I'd like to make. Of course, this would be a BC and we're past the beta phase, so
* are you ok with this change at this point or 
* would you postpone it until the next major version or 
* wouldn't bother about it at all?

lahma:
> "Sensible [syntax tree format](https://github.com/estree/estree/blob/master/es5.md?rgh-link-date=2023-04-09T14%3A57%3A24Z) which follows the standard established by the [ESTree project](https://github.com/estree/estree)".

Sounds good.

>  This is the other change I'd like to make. Of course, this would be a BC and we're past the beta phase, so

I'm team "ok with this change at this point". The idea with next major release is "to break as much as we can to get things cleaned up". I don't expect it's a break the world situation if binary compatibility is broken between RC and release. Not many bother ticking the box in NuGet explorer to get preview releases...

PR #383, opened 2023-04-09
Improve ESTree compliance

adams85:
This PR aims to improve compliance with the ESTree specification.

Besides what have been discussed [here](https://github.com/sebastienros/esprima-dotnet/issues/382), I made an attempt to address a few further issues:
* In our AST model we don't have separate [`BinaryOperator`/`LogicalOperator`](https://github.com/estree/estree/blob/master/es5.md#binary-operations) and [`UnaryOperator`/`UpdateOperator`](https://github.com/estree/estree/blob/master/es5.md#unary-operations) enums as specified by the standard but have the separation w.r.t. the corresponding node types (`BinaryExpression`/`LogicalExpression` and `UnaryExpression`/`UpdateExpression`). However, there were no checks in place to prevent instantiation of these types with a wrong operator type (e.g. a `LogicalExpression` with a multiplication operator.) Thus, checks have been added to prevent invalid combinations.
* The standard defines special cases (subclasses) for [regex](https://github.com/estree/estree/blob/master/es5.md#regexpliteral) and [bigint](https://github.com/estree/estree/blob/master/es2020.md#bigintliteral) literals. In the case of bigints, I don't think we'd gain anything from a subclass since we don't really have additional data to store (the `bigint` property defined by the standard can be replaced with a simple `literal.BigIntValue.ToString()` call in our current model). However, adding a subclass for the regex case seems beneficial as it's a cleaner design and reduces the memory footprint of the AST a bit. (As a matter of fact, we may also consider changing `RegexValue` and `TemplateElementValue` to readonly struct records to save a few further allocations.) I'm not sure about this change though, its impact on existing codebases might be more than acceptable. Please let me know how do you feel about this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ parser (also popularly known as
 ### Features
 
 - Full support for ECMAScript 2019 ([ECMA-262 10th Edition](http://www.ecma-international.org/publications/standards/Ecma-262.htm))
-- Sensible [syntax tree format](https://github.com/estree/estree/blob/master/es5.md) as standardized by [ESTree project](https://github.com/estree/estree)
+- Sensible [syntax tree format](https://github.com/estree/estree/blob/master/es5.md) which follows the standard established by the [ESTree project](https://github.com/estree/estree)
 - Experimental support for [JSX](https://facebook.github.io/jsx/), a syntax extension for [React](https://facebook.github.io/react/)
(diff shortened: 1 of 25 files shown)


lahma (approved):
LGTM, tested migrating Jint and changes were starightforward. Tests green on that side too.

PR #386, opened 2023-05-04
Improve identifier parsing

adams85:
Improves identifier parsing to match the spec even in the astral planes (> \UFFFF) based on Unicode 15.0.

Also, does a minor optimization in the handling of the BMP range (\u0000..\uFFFF):

## Esprima.Benchmark.FileParsingBenchmark

| **Diff**|Method|FileName|Mean|Error|Allocated|
|------- |-------|-------|-------:|-------|-------:|
| Old |ParseProgram|angular-1.2.5|8,474.6 μs|36.10 μs|4293.9 KB|
| **New** |	|	| **8,345.3 μs (-2%)** | **80.26 μs** | **4293.9 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|1,140.7 μs|8.73 μs|673.02 KB|
| **New** |	|	| **1,138.9 μs (0%)** | **17.30 μs** | **673.02 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|6,661.2 μs|44.92 μs|3744.63 KB|
| **New** |	|	| **6,648.2 μs (0%)** | **40.23 μs** | **3744.63 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|10,209.9 μs|19.24 μs|5762.71 KB|
| **New** |	|	| **10,034.3 μs (-2%)** | **60.22 μs** | **5762.71 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|5,427.3 μs|31.40 μs|3074.04 KB|
| **New** |	|	| **5,300.6 μs (-2%)** | **32.30 μs** | **3074.04 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|974.3 μs|6.75 μs|568.59 KB|
| **New** |	|	| **938.8 μs (-4%)** | **6.16 μs** | **568.59 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|4,804.8 μs|7.93 μs|2827.59 KB|
| **New** |	|	| **4,667.7 μs (-3%)** | **18.89 μs** | **2827.59 KB (0%)** |

PR #387, opened 2023-05-05
Update syntax validation rules to align better with the spec

lahma:
While working with support for class members in Jint I've found some problems when running `eval()` code that uses parser and which expects to certain validation rules to apply. I've made things stricter here and basically breaking some *very* legacy code that might use old constructs under `"use strict"` which should anyway be opt-int decision.

There's new options flag `AllowReturnOutsideFunction` which ensures that Jint can use "sloppy test code" but by default return outside of functions is not allowed like the spec mandates.


lahma:
Hey @adams85 , I would need to get this released to complete my class implementation work on Jint side. If we'd release a new RC is there something you would like to get into that train?

adams85:
Hi! Yep, it'd be nice to fit #386 in if you could verify that's ok with Jint. It should be, just makes the parser stricter so we'd better check that it doesn't break anything. However, it's not something critical so if it's urgent for you to make the release, we can as well postpone merging my PR.

PR #390, opened 2023-07-28
Minor improvements

adams85:
Some more polishing.

### DIFF
--- a/src/Esprima/Character.cs
+++ b/src/Esprima/Character.cs
@@ -80,4 +80,23 @@ public static partial class Character
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static bool IsInRange(char c, char min, char max) => c - (uint) min <= max - (uint) min;
+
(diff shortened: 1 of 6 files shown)


lahma (approved):
LGTM 👍🏻

adams85:
Thanks for the quick review!

I had doubts about https://github.com/sebastienros/esprima-dotnet/pull/390/commits/22b6471c6d8b5147550b1b7c15650262d83eaa87, so did some research about exactly how .NET deals with string constants. 

The pre-generated version would result in `ldstr` operations, which works like [this](https://stackoverflow.com/a/31088610/8656352) under the hood. Not quite what I hoped for... So I'd rather drop this change because it'd just increase the assembly size by 3 kB without a real benefit.

adams85:
I believe the time will be ripe for a new RC (hopefully the last one before the final 3.x). If you agree, would you do the honors, please? :)

lahma:
[3.0.0-rc-03](https://www.nuget.org/packages/Esprima/3.0.0-rc-03) published! 🚀

Issue #322, opened 2022-08-19
Optional chaining-related parser issues


adams85:
The following snippet is parsed in non-tolerant mode whereas it should report an error like "Uncaught SyntaxError: Invalid optional chain from new expression"

```js
var a = { b: function f() { return this; } };
new a?.b; // correct syntax: new (a?.b)
new a?.b(x); // correct syntax: new (a?.b)(x)
```

Similar constructs like tagged templates are handled correctly ("Invalid tagged template on optional chain" is reported when parentheses omitted) but not decorators:

```js
@a?.b class X { } // correct syntax: @(a?.b) class X { }
```

PR #393, opened 2023-08-06
Fix parsing of optional chaining in new expressions

adams85:
Takes a crack at https://github.com/sebastienros/esprima-dotnet/issues/322 (However, addresses only new expressions because [decorators seem to have bigger issues](#394) ATM).

The fix involves [keeping track of the last processed token](https://github.com/sebastienros/esprima-dotnet/compare/fix/optional-chaining-in-new?expand=1#diff-92a67081d35facbe8921de0638d8eccd7f7204c8b99a656abc3af2ea750857c1L340), which I'm not completely comfortably with TBH. But other approaches look much more complicated. Could someone more familiar with the parser take a look at this?


lahma:
Now sure if I can be of much help here, always appreciate you improving the parsing though! I've usually looked at [acorn](https://github.com/acornjs/acorn) for ideas on how to accomplish some specific parsing scenarios in the past.

adams85:
Thanks for the tip. Acorn just passes a flag down the call stack to implement the necessary check. In Esprima it's a bit more complicated because `ParseLeftHandSideExpression` is not called directly but through `IsolateCoverGrammar`. So I had to introduce a new flag in `JavaScriptParser.Context`, which I wanted to avoid. However, this looks to be the correct way.

And while being at it, I also fixed #394.

lahma (approved):
LGTM, seems naturally follow common patterns like previous-reset for flags and reads well. Nicely done once again!

adams85:
How about pushing out a new RC? :) If you could do it, I could finish the regex story in Jint.

lahma:
I of course can, but would you like to do the honors this time? Just to make sure we don't have dependency on single entities (like me or Sebastien). Just create the tag with the name following earlier tags against latest main, push against origin and rock'n'roll 🚀 

Just say so and I'll do it, just testing how you would feel about it (convo about whether to push is of course nice, but you master the features at the moment).

adams85:
If it's that simple, no problem. :) I'll give it a shot.

lahma:
What is this sorcery, you have accomplished the task of publishing an Esprima release to nuget.org, congrats! 🚀

PR #396, opened 2023-08-10
Improve ArrayList and NodeList

lahma:
I was checking how much faster it is to use span instead of indexing (Jint uses indexing at the moment) and I though that it shouldn't have that much difference as it had so tweaked code a bit.

* better indexer with less IL - it's enough to have `nameof(something)` to cause `ldstring`, more fine-grained ThrowHelper to the rescue
* encapsulate `EnsureCapacity` as single method to resize
* remove `TrimExcess` as I didn't see much point in it performance wise

# Results

Modified benchmark (no boxing for result) was used against both `main` and this PR. The couple byte allocations in 0 size test case ForLoopBenchmark is quite curious, but didn't find out the reason. Now indexing is as fast as spans or even faster for small lists (0-1), spans start to shine when iterating more items.

💡 We maybe should consider converting visitors to use foreach + span.

```

BenchmarkDotNet v0.13.7, Windows 11 (10.0.23516.1000)
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK 7.0.400
  [Host]     : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2


```

## Esprima.Benchmark.AstTreeWalkBenchmark

| Old |VisitChildren|2.565 ms|0.0151 ms|35 B|
| **New** |	| **2.572 ms (0%)** | **0.0140 ms** | **35 B (0%)** |

| Old |ParseProgram|angular-1.2.5|10.231 ms|0.0621 ms|4095.44 KB|
| **New** |	|	| **10.238 ms (0%)** | **0.0449 ms** | **4095.76 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|1.348 ms|0.0018 ms|647.77 KB|
| **New** |	|	| **1.329 ms (-1%)** | **0.0083 ms** | **647.8 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|7.932 ms|0.0374 ms|3540.14 KB|
| **New** |	|	| **7.997 ms (+1%)** | **0.0368 ms** | **3541.17 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|12.138 ms|0.0350 ms|5609.22 KB|
| **New** |	|	| **12.053 ms (-1%)** | **0.0513 ms** | **5609.56 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|6.363 ms|0.0213 ms|2938.37 KB|
| **New** |	|	| **6.377 ms (0%)** | **0.0284 ms** | **2938.73 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|1.156 ms|0.0048 ms|555.5 KB|
| **New** |	|	| **1.139 ms (-1%)** | **0.0046 ms** | **555.45 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|5.439 ms|0.0223 ms|2714.65 KB|
| **New** |	|	| **5.476 ms (+1%)** | **0.0237 ms** | **2714.93 KB (0%)** |

## Esprima.Benchmark.NodeListEnumerationBenchmark

| **Diff**|Method|NodeListIndex|Mean|Error|Allocated|
| Old |For_DirectIndexing|0|5,633,143.177 ns|25,214.8469 ns|7 B|
| **New** |	| **0** | **2,318,609.390 ns (-59%)** | **11,763.7113 ns** | **3 B (-57%)** |
| Old |For_SpanIndexing|0|2,629,088.965 ns|3,468.6020 ns|3 B|
| **New** |	| **0** | **2,619,675.417 ns (0%)** | **8,738.2733 ns** | **4 B (+33%)** |
| Old |ForEach_Span|0|2,630,783.750 ns|7,398.6263 ns|3 B|
| **New** |	| **0** | **2,698,511.576 ns (+3%)** | **50,837.2970 ns** | **3 B (0%)** |
| Old |For_DirectIndexing|1|1,782.627 ns|1.1875 ns|0 B|
| **New** |	| **1 (0%)** | **685.421 ns (-62%)** | **2.2029 ns** | **0 B** |
| Old |For_SpanIndexing|1|669.814 ns|3.1510 ns|0 B|
| **New** |	| **1 (0%)** | **673.555 ns (+1%)** | **0.1979 ns** | **0 B** |
| Old |ForEach_Span|1|674.983 ns|0.6107 ns|0 B|
| **New** |	| **1 (0%)** | **674.960 ns (0%)** | **4.5344 ns** | **0 B** |
| Old |For_DirectIndexing|2|20.300 ns|0.1101 ns|0 B|
| **New** |	| **2 (0%)** | **5.302 ns (-74%)** | **0.0018 ns** | **0 B** |
| Old |For_SpanIndexing|2|4.746 ns|0.0197 ns|0 B|
| **New** |	| **2 (0%)** | **4.704 ns (-1%)** | **0.0063 ns** | **0 B** |
| Old |ForEach_Span|2|4.799 ns|0.0218 ns|0 B|
| **New** |	| **2 (0%)** | **4.735 ns (-1%)** | **0.0129 ns** | **0 B** |

| Old |Visit|11.54 ms|0.225 ms|14 B|
| **New** |	| **10.49 ms (-9%)** | **0.103 ms** | **14 B (0%)** |
| Old |Rewrite|23.60 ms|0.467 ms|28 B|
| **New** |	| **22.46 ms (-5%)** | **0.157 ms** | **28 B (0%)** |

### DIFF
--- a/samples/Esprima.Benchmark/NodeListEnumerationBenchmark.cs
+++ b/samples/Esprima.Benchmark/NodeListEnumerationBenchmark.cs
@@ -3,10 +3,8 @@ using Esprima.Ast;
 
 namespace Esprima.Benchmark;
 
-[MemoryDiagnoser]
 [RankColumn]
-[CsvExporter]
-[MarkdownExporterAttribute.GitHub]
+[MemoryDiagnoser]
 public class NodeListEnumerationBenchmark
 {
     private const string FileName = "bundle";
@@ -30,7 +28,7 @@ public class NodeListEnumerationBenchmark
     }
 
     [Benchmark]
-    public object For_DirectIndexing()
+    public int For_DirectIndexing()
     {
         var nodeList = _nodeLists[NodeListIndex];
 
@@ -43,7 +41,7 @@ public class NodeListEnumerationBenchmark
     }
 
     [Benchmark]
-    public object For_SpanIndexing()
+    public int For_SpanIndexing()
     {
         var nodeList = _nodeLists[NodeListIndex].AsSpan();
 
@@ -56,7 +54,7 @@ public class NodeListEnumerationBenchmark
     }
 
     [Benchmark]
-    public object ForEach_Span()
+    public int ForEach_Span()
     {
         var nodeList = _nodeLists[NodeListIndex].AsSpan();
 
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -80,7 +80,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
     {
         if (initialCapacity < 0)
         {
-            ThrowArgumentOutOfRangeException<T>(nameof(initialCapacity), initialCapacity, null);
+            ThrowInvalidInitialCapacity();
         }
 
         _items = initialCapacity > 0 ? new T[initialCapacity] : null;
@@ -90,6 +90,12 @@ internal struct ArrayList<T> : IReadOnlyList<T>
         _localVersion = 0;
         _sharedVersion = null;
 #endif
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void ThrowInvalidInitialCapacity()
+        {
+            ThrowArgumentException("Invalid initial capacity", nameof(initialCapacity));
+        }
     }
 
     /// <remarks>
@@ -108,6 +114,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
 
     public int Capacity
     {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         get
         {
             AssertUnchanged();
@@ -117,27 +124,12 @@ internal struct ArrayList<T> : IReadOnlyList<T>
         {
             AssertUnchanged();
 
-            if (value < _count)
+            if (value < (_items?.Length ?? 0))
             {
-                ThrowArgumentOutOfRangeException<T>(nameof(value), value, null);
+                ThrowArgumentOutOfRangeException(nameof(value), value, null);
             }
-            else if (value == (_items?.Length ?? 0))
-            {
-                return;
-            }
-            else if (value > 0)
-            {
-                T[] array = new T[value];
-                if (_count > 0)
-                {
-                    Array.Copy(_items, 0, array, 0, _count);
-                }
-                _items = array;
-            }
-            else
-            {
-                _items = null;
-            }
+
+            EnsureCapacity(value);
 
             OnChanged();
         }
@@ -166,7 +158,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
                 return _items![index];
             }
 
-            return ThrowArgumentOutOfRangeException<T>(nameof(index), index, null);
+            return ThrowIndexOutOfRangeException<T>();
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -181,7 +173,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
                 return;
             }
 
-            ThrowArgumentOutOfRangeException<T>(nameof(index), index, null);
+            ThrowIndexOutOfRangeException<T>();
         }
     }
 
@@ -221,10 +213,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
         var oldCount = _count;
         var newCount = oldCount + itemCount;
 
-        if (Capacity < newCount)
-        {
-            Array.Resize(ref _items, Math.Max(newCount, MinAllocatedCount));
-        }
+        EnsureCapacity(newCount);
 
         Debug.Assert(_items is not null);
         items.CopyTo(_items.AsSpan(oldCount, itemCount));
@@ -242,12 +231,7 @@ internal struct ArrayList<T> : IReadOnlyList<T>
     {
         AssertUnchanged();
 
-        var capacity = Capacity;
-
-        if (_count == capacity)
-        {
-            Array.Resize(ref _items, Math.Max(capacity * 2, MinAllocatedCount));
-        }
+        EnsureCapacity(_count + 1);
 
         Debug.Assert(_items is not null);
         _items![_count] = item;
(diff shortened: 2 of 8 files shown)


src/Esprima/ArrayList.cs
```diff
@@ -348,16 +327,6 @@ public void Yield(out T[]? items, out int count)
         this = default;
     }
 
-    public void TrimExcess(int threshold = MinAllocatedCount)
```
adams85, line 351: The poor thing didn't live long... :(

I added it not to improve performance but to reduce memory footprint in cases where we build a list with `ArrayList<T>`, then hand it over to the consumer via a public API as a `ReadOnlySpan<T>` (to avoid copying the list in lucky cases where the underlying array is almost full, that is, doesn't have many unused items).

For example, in the case of collecting comments in the scanner, it could happen that there are 1025 comments in the scanned script, so the underlying array's capacity will increase to 2048. Which means that there will be 1023 unused items, which will be kept around in the memory since we return the underlying array via a span.

`TrimExcess` was introduced to improve this situation, so I'd still keep it.

src/Esprima/JavascriptParser.cs
```diff
@@ -50,7 +50,7 @@ public void ReleaseLargeBuffers()
             Decorators.Clear();
             if (Decorators.Capacity > 64)
             {
-                Decorators.Capacity = 64;
```
adams85, line 53: This change is not necessary if `Capacity` worked as I suggested.

src/Esprima/Scanner.RegExpParser.cs
```diff
@@ -185,7 +185,6 @@ public RegExpParseResult Parse()
             }
 
             Debug.Assert(conversionError is null);
-            capturingGroups.TrimExcess();
```
adams85, line 188: IMO we'd better trim the collected `capturingGroups` one way or the other because it's exposed in the public API (see `RegExpParseResult`).

src/Esprima/Scanner.cs
```diff
@@ -144,7 +144,7 @@ internal void ReleaseLargeBuffers()
         _curlyStack.Clear();
         if (_curlyStack.Capacity > 16)
         {
-            _curlyStack.Capacity = 16;
```
adams85, line 147: This change is not necessary if `Capacity` worked as I suggested.

src/Esprima/Scanner.cs
```diff
@@ -517,7 +517,6 @@ internal ArrayList<Comment> ScanCommentsInternal()
     public ReadOnlySpan<Comment> ScanComments()
     {
         var comments = ScanCommentsInternal();
-        comments.TrimExcess();
```
adams85, line 520: IMO we'd better trim the collected `comments` one way or the other because it's exposed in the public API.

However, if you don't see the point in reducing memory footprint and don't want to remove excess, then `ScanCommentsInternal` is unnecessary.

src/Esprima/ArrayList.cs
```diff
@@ -398,6 +367,18 @@ IEnumerator IEnumerable.GetEnumerator()
         return GetEnumerator();
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void EnsureCapacity(int capacity)
+    {
+        var temp = _items;
+        if (temp == null || temp.Length < capacity)
+        {
+            var minAllocatedCount = temp == null || temp.Length < MinAllocatedCount ? MinAllocatedCount : temp.Length * 2;
```
adams85, line 376: nit:

```suggestion
        if (temp is null || temp.Length < capacity)
        {
            var minAllocatedCount = temp is null || temp.Length < MinAllocatedCount ? MinAllocatedCount : temp.Length * 2;
```

samples/Esprima.Benchmark/NodeListEnumerationBenchmark.cs
```diff
@@ -30,7 +28,7 @@ public void Setup()
     }
 
     [Benchmark]
-    public object For_DirectIndexing()
```
adams85, line 33: Damn, nice oversight on my part... This pretty much messed up the benchmark results.

Nice catch! 👍

adams85:
> Now indexing is as fast as spans or even faster for small lists (0-1), spans start to shine when iterating more items.
>
> 💡 We maybe should consider converting visitors to use foreach + span.

Sounds like it's worth a try. Thanks to our source generation magic, probably it wouldn't be too tedious. So, I'm looking forward to the new benchmark results! ;)

src/Esprima/ArrayList.cs
```diff
@@ -117,28 +124,13 @@ public int Capacity
...
-                }
-                _items = array;
-            }
-            else
-            {
-                _items = null;
+                ThrowArgumentOutOfRangeException(nameof(value), value, null);
             }
 
+            EnsureCapacity(value);
```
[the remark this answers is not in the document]
lahma: Fair enough, I'll revert the capacity related changes.

src/Esprima/ArrayList.cs
lahma: Reverted the removal.

src/Esprima/ArrayList.cs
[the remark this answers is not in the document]
adams85: I had problems with the `Capacity` property only, `EnsureCapacity` seemed ok in other places. But I'm also ok with the original logic.

adams85:
FWIW, [`ReadOnlySpan<T>` ctor](https://github.com/dotnet/runtime/blob/v6.0.21/src/libraries/System.Private.CoreLib/src/System/ReadOnlySpan.cs#L61) does one less checks than [`Span<T>` ctor](https://github.com/dotnet/runtime/blob/v6.0.21/src/libraries/System.Private.CoreLib/src/System/Span.cs#L64).

I assume that the worse performance for smaller lists is caused by the initial cost of constructing a span. So, for read-only iteration `ReadOnlySpan<T>` is a tiny bit better.

adams85 (approved):
Those indexer perf gains are serious. Kudos!

lahma:
Thanks for helping out and getting this finalized!

PR #397, opened 2023-08-10
Replace custom attribute shims with PolySharp

lahma:
While the original reasoning to have own versions of the attribute shims due to concerns about supply chain attacks, I would argue that we can trust in the author of [PolySharp](https://github.com/Sergio0694/PolySharp) who is a well recognized Microsoft employee. This would also gives us many other missing bits if we ever needed them. Thoughts?


adams85 (approved):
As long as it's a dev-time dependency, I don't have too strong concerns for adding a 3rd party dependency. The lib and its author look solid, so I have nothing against it.

I just suggest we are explicit about the included polyfills, that is, let's include only those which are really needed. (Sorry for meddling with your PR, was quicker this way for me. Just drop my commit if you disagree.)

Issue #303, opened 2022-08-05
Eliminate JavaScript letter casing inconsistencies


adams85:
Unfortunately, the AST to JS feature was implemented with an inconsistent letter casing: "Javascript" was used in code and file names instead of "JavaScript". The confusion probably came from the parser class: it's named "JavaScriptParser" but it resides in a file named "JavascriptParser.cs".

Anyway, this inconsistency should be addressed before releasing v3. I think we'd better stick to the letter casing "JavaScript". (This one is used in .NET BCL too: e.g. `HttpUtility.JavaScriptStringEncode` or `JavaScriptEncoder`.)

So we have the following tasks:
* [x] Correct the letter casing in the code.
* [ ] Correct the letter casing in the file names. (This looks however somewhat problematic, especially in the case of *JavascriptParser.cs*, as it would probably cause merge conflicts and possibly mess up git blame. So if we want to do this, we'd better do it right before the production release, after everything else has been done for v3.)

jogibear9988:
As I'm developing on Windows, FileName Casing does not matter, wich is often a problem. So normaly I rename the file to a complete new name, commit, and then rename it again to the new wished one

jogibear9988:
close?

adams85:
Nope, file names should be made consistent with the names of classes they contain if that doesn't mess up git blame (much).

> So normaly I rename the file to a complete new name, commit, and then rename it again to the new wished one

I've tested it and it looks like git is case-sensitive, thus, it allows us to do this in one pass. E.g. you can rename *JavascriptParser.cs* to *JavaScriptParser.cs*.

adams85:
Nope, file names should be made consistent with the names of classes they contain if that doesn't mess up git blame (much).

> So normaly I rename the file to a complete new name, commit, and then rename it again to the new wished one

I've tested it and it looks like git is case-sensitive, thus, it allows us to do this in one pass. E.g. you can rename `JavascriptParser.cs` to `JavaScriptParser.cs`.

jogibear9988:
yes, but i think then a checkout on windows could produce problems. (we had some)

jogibear9988:
cause i think i windows you could not rename only caseing changes

jogibear9988:
but i could be wrong

lahma:
On Windows, you need to use git command line to do case-only change: https://stackoverflow.com/questions/1793735/change-case-of-a-file-on-windows#1793757

adams85:
[...]

PR #398, opened 2023-08-18
Fix number parsing issues

adams85:
Fixes several bugs in number parsing (especially bigints):
* `0o` bigint literals should be parsed as octal numbers (were incorrectly parsed as decimal).
* `0b` and `0o` literals which are larger than a 64-bit integer are valid and should be parsed.
* `0x<256 or more zeros>` should not result in `double.PositiveInfinity` (see https://github.com/sebastienros/esprima-dotnet/compare/fix/number-parsing?expand=1#diff-5c649e707747c206cd95c805966b187aa0e3f0725f25f51aef684f8254ad0149L1065).
* Legacy octal numbers with group separators (e.g. `0777_777`) should be a syntax error (`FormatException` was thrown).
* Fractions with omitted leading zero and trailing 'n'  (e.g. `.17n`) should be a syntax error (no exception was thrown).

Also fixes a few minor bugs (e.g. in `AstToJsonConverter` when numeric literal value is `Infinity`).

&plus; Some micro-optimization.

### DIFF
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -370,13 +370,18 @@ internal struct ArrayList<T> : IReadOnlyList<T>
     /// </remarks>
     public Span<T> AsSpan() => new Span<T>(_items, 0, _count);
 
+    /// <remarks>
+    /// Items should not be added or removed from the <see cref="ArrayList{T}"/> while the returned <see cref="ReadOnlySpan{T}"/> is in use!
+    /// </remarks>
+    public ReadOnlySpan<T> AsReadOnlySpan() => new ReadOnlySpan<T>(_items, 0, _count);
+
 #if NETSTANDARD2_1_OR_GREATER
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
 #endif
     public T[] ToArray()
     {
 #if NETSTANDARD2_1_OR_GREATER
-        return AsSpan().ToArray();
+        return AsReadOnlySpan().ToArray();
 #else
         if (_count == 0)
         {
--- a/src/Esprima/ArrayList.cs
+++ b/src/Esprima/ArrayList.cs
@@ -368,11 +368,13 @@ internal struct ArrayList<T> : IReadOnlyList<T>
     /// <remarks>
     /// Items should not be added or removed from the <see cref="ArrayList{T}"/> while the returned <see cref="Span{T}"/> is in use!
     /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public Span<T> AsSpan() => new Span<T>(_items, 0, _count);
 
     /// <remarks>
     /// Items should not be added or removed from the <see cref="ArrayList{T}"/> while the returned <see cref="ReadOnlySpan{T}"/> is in use!
     /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public ReadOnlySpan<T> AsReadOnlySpan() => new ReadOnlySpan<T>(_items, 0, _count);
 
 #if NETSTANDARD2_1_OR_GREATER
--- a/src/Esprima/Scanner.cs
+++ b/src/Esprima/Scanner.cs
@@ -418,7 +418,7 @@ public sealed partial class Scanner
                     var comment = SkipSingleLineComment(2);
                     if (_trackComment)
                     {
-                        comments.AddRange(comment.AsSpan());
+                        comments.AddRange(comment.AsReadOnlySpan());
                     }
 
                     start = true;
@@ -430,7 +430,7 @@ public sealed partial class Scanner
                     var comment = SkipMultiLineComment();
                     if (_trackComment)
                     {
-                        comments.AddRange(comment.AsSpan());
+                        comments.AddRange(comment.AsReadOnlySpan());
                     }
                 }
                 else
@@ -449,7 +449,7 @@ public sealed partial class Scanner
                     var comment = SkipSingleLineComment(3);
                     if (_trackComment)
                     {
-                        comments.AddRange(comment.AsSpan());
+                        comments.AddRange(comment.AsReadOnlySpan());
                     }
                 }
                 else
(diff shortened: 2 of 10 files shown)


adams85:

| Old |ParseProgram|angular-1.2.5|8,737.0 μs|122.22 μs|4095.44 KB|
| **New** |	|	| **8,604.0 μs (-2%)** | **781.46 μs** | **4095.47 KB (0%)** |
| Old |ParseProgram|backbone-1.1.0|1,099.7 μs|24.22 μs|647.78 KB|
| **New** |	|	| **1,113.5 μs (+1%)** | **64.11 μs** | **647.78 KB (0%)** |
| Old |ParseProgram|jquery-1.9.1|6,561.9 μs|184.62 μs|3540.13 KB|
| **New** |	|	| **6,597.9 μs (+1%)** | **292.46 μs** | **3540.13 KB (0%)** |
| Old |ParseProgram|jquery.mobile-1.4.2|10,105.1 μs|780.14 μs|5609.22 KB|
| **New** |	|	| **10,184.9 μs (+1%)** | **535.63 μs** | **5609.24 KB (0%)** |
| Old |ParseProgram|mootools-1.4.5|5,262.7 μs|239.04 μs|2938.38 KB|
| **New** |	|	| **5,195.0 μs (-1%)** | **76.77 μs** | **2938.37 KB (0%)** |
| Old |ParseProgram|underscore-1.5.2|953.8 μs|24.72 μs|555.5 KB|
| **New** |	|	| **927.7 μs (-3%)** | **169.11 μs** | **555.5 KB (0%)** |
| Old |ParseProgram|yui-3.12.0|4,521.4 μs|112.76 μs|2714.65 KB|
| **New** |	|	| **4,486.3 μs (-1%)** | **226.24 μs** | **2714.65 KB (0%)** |

src/Esprima/Scanner.cs
```diff
@@ -1162,8 +1180,17 @@ private Token ScanBinaryLiteral(int start)
             }
         }
 
-        var numberString = number.ToString();
-        var value = Convert.ToUInt32(numberString, 2);
+        double value;
+
+        if (number.Length <= 64)
+        {
+            value = Convert.ToUInt64(number.ToString(), fromBase: 2);
```
lahma, line 1187: looking how this is implemented in NET 6.0, would be nice to get access to that `ParseNumbers.StringToLong(ReadOnlySpan<char> s, int radix, int flags)` which is internal...

lahma (approved):
Nice improvements! 👍🏻

src/Esprima/ArrayList.cs
```diff
@@ -370,13 +370,18 @@ public void TrimExcess(int threshold = MinAllocatedCount)
     /// </remarks>
     public Span<T> AsSpan() => new Span<T>(_items, 0, _count);
 
+    /// <remarks>
+    /// Items should not be added or removed from the <see cref="ArrayList{T}"/> while the returned <see cref="ReadOnlySpan{T}"/> is in use!
+    /// </remarks>
+    public ReadOnlySpan<T> AsReadOnlySpan() => new ReadOnlySpan<T>(_items, 0, _count);
```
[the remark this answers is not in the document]
adams85: > this might allow automatic conversion

Hmm, while it may be a good idea in general, in this case I don't think so because the underlying list can be mutated as pointed out in the XML doc. It's preferable that the consumer of this API encounters this message, which they won't if we allow implicit conversion.

> not sure if would be good to sprinkle some MethodImpl(MethodImplOptions.AggressiveInlining) to As*Span methods too, just in case

This is a good idea, however. At least, it can't hurt.

src/Esprima/Scanner.cs
adams85: Yep, this situation if unfortunate. Maybe they'll expose this API in a later version.

src/Esprima/ArrayList.cs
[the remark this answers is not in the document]
lahma: > > this might allow automatic conversion
> 
> Hmm, while it may be a good idea in general, in this case I don't think so because the underlying list can be mutated as pointed out in the XML doc. It's preferable that the consumer of this API encounters this message, which they won't if we allow implicit conversion.

A very good point. Whilst the type is internal, anyone could do that mistake. I guess my idea would still apply for `NodeList` as that one is read-only 🤔

src/Esprima/Scanner.cs
lahma: Actually looking at the current code, we could probably just copy-paste it and un-localize the error messages and we would have the solution. Not sure if it's worth it though.

https://github.com/dotnet/runtime/blob/6cd25fbeb1a2a332561a46cdf2b01d06c1fbb192/src/libraries/System.Private.CoreLib/src/System/ParseNumbers.cs

lahma:
@adams85  a bit unrelated, but what are your thoughts about 3.0 maturity, should the big release be made and leave smaller bug fixes to maintenance releases. Or do you see some API work ahead still?

src/Esprima/ArrayList.cs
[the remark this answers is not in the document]
adams85: Yep, `NodeList` looks like a better candidate for your idea, however I don't think we'd gain much even in this case because when `NodeList`s are accessed, they are mostly enumerated in `for` or `foreach` loops. E.g.:

```cs
foreach (var node in nodeList)
{
   /* ... */
}
```

Even though we define an implicit conversion, the standard enumerator will be used instead of the ROS enumerator. For that you'd need to type out the explicit type cast to ROS, create a ROS variable before the loop, etc. Which looks more cumbersome than just snapping an `.AsSpan()` to the end of `nodeList`...

src/Esprima/Scanner.cs
adams85: Not sure either but it's very likely that there's further room for improvement in this area. Currently I don't plan to spend more time on it but feel free to experiment with this if you .

adams85:
Nope, as far as I'm concerned, no new features are planned. IMO the parser is pretty solid now, I only know of one area (syntax error detection and reporting) which would need some love but that would take a lot of work for not pretty little gain. I can't see that work be done soon (at least, in the current codebase - but more on this later.) So, I don't think that minor bugs and refinements should block 3.0.

There are a few cleanup tasks in my todo list like [this](https://github.com/sebastienros/esprima-dotnet/issues/303). Hopefully, I can take care of them in the next week(s). I'll add those tasks to the [3.0 milestone](https://github.com/sebastienros/esprima-dotnet/milestone/1) so we can keep track of the progress. I suggest you do the same if you have anything in mind that you'd like to make its way into 3.0

PR #402, opened 2023-08-21
Code cleanup before v3.0 final release

adams85:


lahma (approved):

adams85:
Thanks for the review. As for me, this is the last finishing touch. Do you have anything else in mind that you'd like to include in v3.0 or can we proceed with the release?

If there's nothing else, shall I "do the honors"?

lahma:
I believe we are good to go. The honors are most definitely yours as you are pretty much the force that pushed the work needed for this big release. 

We are really grateful for all the work you've done! 🚀

PR #410, opened 2023-09-10
Add strict option to ParseExpression

adams85:


src/Esprima/JavaScriptParser.cs
```diff
@@ -2751,11 +2751,12 @@ private Expression ParseExpression()
     /// <summary>
     /// Parses the code as a JavaScript expression.
     /// </summary>
-    public Expression ParseExpression(string code)
+    public Expression ParseExpression(string code, bool strict = false)
```
lahma, line 2754: nit: this is a binary breaking change compared to another overload - from 3.0.1 servicing perspective. I don't mind it but a bit against semantic versioning

adams85: You're absolutely right. What's more, the [previous PR introduced a similar ABI BC](https://github.com/sebastienros/esprima-dotnet/commit/423d6911812a7c5596bdfcd2f410f2160fde8854#diff-20bf466efe8eabea6e3dec4f4876432fc19a60d4499e7703e2fdae1d48d255bfR122) (pretty small expected impact though).

What do you suggest, how should we deal with this kind of minor, binary-level BCs which cause no source-level BCs? Strict semver would dictate v4.0.0 but that would feel a bit of a stretch... How about a v3.0.1 (or v3.1.0 maybe?) + mentioning the BC in the release notes?

lahma: My hunch is that people either directly depend on Esprima (they just update the reference) or it's a transitive dependency (like Jint, they reference Jint and are happy with whatever comes with it as long as it works).

So this would affect scenarios where you have explicit references for both Jint and Esprima and would be eager to update Esprima as soon as new version available.

There's no explicit promise for semantic versionin AFAIK so if I'd just push a 3.0.1 with these (and possible other) changes at some point. Maybe outcry about bad versioning will later help steer the decisions 😉

adams85: > So this would affect scenarios where you have explicit references for both Jint and Esprima and would be eager to update Esprima as soon as new version available.

I did a quick search in Jint codebase and found no calls to `ParseExpression`. So it seems this BC can't lead to issues anyway.

> There's no explicit promise for semantic versionin AFAIK so if I'd just push a 3.0.1 with these (and possible other) changes at some point. Maybe outcry about bad versioning will later help steer the decisions 😉
👍

lahma: Those release notes you've published should be more than enough (kudos for that!). I as a consumer of the library I would be more than satisfied to find such detailed warnings about binary compatibility. Of course I might not be the usual consumer as I read release notes 😉

PR #413, opened 2023-10-12
Fix a few edge cases in AST to JS conversion

### LINKED ISSUE #411: Issues with optional chaining
```js
const user = {
  profile: {
    name: 'Alice',
    address: {
      city: 'Wonderland'
    }
  }
};

const city = user?.profile?.address?.city;
```

It doesn't seem to like this code, with the dot following the question mark. It is valid JS from what I can tell, with optional chaining being introduced in ES2020.

adams85:
Makes changes to `AstToJsonConverter` to correctly parenthesize logical expressions which mix nullish coalescing and logical AND/OR operations (e.g. `a ?? b && c`).

Also, fixes formatting of plus/minus binary operations having a right side that starts with an ambiguous punctuator (e.g. `a+++b` vs `a+ ++b`).

### DIFF
--- a/src/Esprima/Ast/Literal.cs
+++ b/src/Esprima/Ast/Literal.cs
@@ -7,9 +7,6 @@ namespace Esprima.Ast;
 [VisitableNode(SealOverrideMethods = true)]
 public partial class Literal : Expression
 {
-    private static readonly object s_boxedTrue = true;
-    private static readonly object s_boxedFalse = false;
-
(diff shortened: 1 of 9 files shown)


adams85:
Setting the status to draft as I discovered a few other problematic cases:
* `a/ /regex/`
* `a<! --b`
* `a-- >b`

These are converted to JS incorrectly when formatting is turned off. That is, the PR needs some more work to handle these as well.

lahma (approved):
LGTM as far as I can tell, the test suite gives me confidence 👍🏻

PR #417, opened 2023-11-03
Fix expression parsing in export default statements

### LINKED ISSUE #416: Issue with parsing spread operator
**Version used**

3.0.1

**Describe the bug**

I'm running into a case where the parser doesn't seem to like my spread operator. I've had a hell of a time trying to narrow spread operator issues down, as I've observed multiple. But they all appear to be non-trivial to reduce to a simple reproduction example. 

**To Reproduce**

```csharp

    [Fact]
    public void TestDotExport()
    {
        var javascriptCode = @"

export default [
  
].map((e, i) => ({
    ...e,
    index: i,
}));


";

        var parsed = new JavaScriptParser(new ParserOptions { Tolerant = true });
        var node = parsed.ParseModule(javascriptCode);
        node.Should().NotBeNull();
        var prettied = node.ToJavaScriptString(true);
        node = parsed.ParseModule(prettied);
        node.Should().NotBeNull();
    }
```

```
Line 5: Unexpected token .
   at Esprima.JavaScriptParser.ThrowUnexpectedToken(Token& token, String message)
   at Esprima.JavaScriptParser.ConsumeSemicolon()
   at Esprima.JavaScriptParser.ParseExportDeclaration()
   at Esprima.JavaScriptParser.ParseStatementListItem()
   at Esprima.JavaScriptParser.ParseModule(String code, String source)

```

**Expected behavior**
This should parse without issues.

adams85:
Fixes #416

See also https://tc39.es/ecma262/#prod-ExportDeclaration

### DIFF
--- a/src/Esprima/JavaScriptParser.cs
+++ b/src/Esprima/JavaScriptParser.cs
@@ -5421,9 +5421,7 @@ ParseValue:
                 // export default {};
                 // export default [];
                 // export default (1 + 2);
-                var declaration = Match("{")
-                    ? ParseObjectInitializer()
-                    : Match("[") ? ParseArrayInitializer() : ParseAssignmentExpression();
+                var declaration = ParseAssignmentExpression();
 
                 ConsumeSemicolon();
                 exportDeclaration = Finalize(node, new ExportDefaultDeclaration(declaration));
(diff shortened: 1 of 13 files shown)


lahma (approved):
The issue's test case seems to pass now. Should the test case be added to test suite too (I'm also fine with adding FluentAssertions as dependency as xUnit asserts are yucky anywa)?

adams85:
> Should the test case be added to test suite too

Absolutely, I just forgot to add it... :) Thanks for pointing this out.

> I'm also fine with adding FluentAssertions as dependency as xUnit asserts are yucky anywa

Unfortunately, I'm not familiar with FluentAssertions but I don't think we'd need it for this one. However, feel free to add it if you see it useful.

PR #419, opened 2023-11-13
Prepare for NET 8 analyzers

lahma:
Another set of tweaks that NET 8 analyzers complain about.


src/Esprima/Ast/SyntaxComment.cs
```diff
@@ -26,7 +28,7 @@ public override string ToString()
                     writer.WriteLineComment(Value, JavaScriptTextWriter.TriviaFlags.None);
                     break;
                 case CommentType.Block:
-                    writer.WriteBlockComment(Value.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None), JavaScriptTextWriter.TriviaFlags.None);
```
sebastienros, line 29: 😅

src/Esprima/ParserExtensions.cs
```diff
@@ -198,6 +198,14 @@ internal static bool Contains(this string source, char c)
     }
 #endif
 
+#if NETFRAMEWORK || NETSTANDARD2_0
```
sebastienros, line 201: Note on newer versions it uses vectorization (you know). We might want to use these and IndexOf as much as possible. I believe it's a jit thing though so it shouldn't require custom tfms.

lahma: Yes I think the interesting case is to detect whether string contains something before doing replace, like in https://github.com/RicoSuter/NJsonSchema/pull/1638/files . It's not ideal with current TFMs, but maybe [SearchValues](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.searchvalues-1?view=net-8.0) would fit the bill nicely - just to check something before doing and expensive chain of replaces.

PR #432, opened 2024-01-31
Update README.md: fix Build badge

davidnx:
React to #431 which renamed staging.yml to build.yml and broke the badge on the project start page.


lahma (approved):
Good catch, thank you!

sebastienros:
@lahma FYI David is one of the most brilliant person I have worked with at MS. You should be as proud as I am he's using Jint ;)

lahma:
I'm just standing on the shoulders of the giants and pushing the merge button 😄 

Great to hear that the project has some use  👍🏻

PR #443, opened 2026-04-14
Fix bugs in regexp conversion

adams85:
Backports fixes for some recently discovered nasty bugs in JS RegExp-to-.NET Regex conversion from Acornima:
- Fixes conversion of `\w` and `\W` character classes in regexp patterns. (U+0130 should not be matched.)
- Fixes conversion of the `$` anchor in regexp patterns. (In .NET, `\z` should be used instead of `$` to avoid matching before potential newline at the end of the input string.)
- Fixes validation of regexp patterns having a non-escaped backslash at the end.

Even though the project is no longer active, it might still be used, so I believe these blatant mistakes should be corrected.

BTW, what do you think, is the time ripe for some kind of deprecation message like "This project is not actively maintained anymore. If you need an up-to-date ECMAScript parser, you can try Acornima."?


adams85:
Well, I won't bother with the deprecation message then, will just try to release the regexp fixes.

